PR #1, opened 2017-04-05
Service for gathering and reporting metrics

Scooletz:
Connects to https://github.com/Particular/LaunchWave.DevOpsAdvancedMonitoring/issues/48

This PR implements the Program Service that:
- [x] uses bootstrapper to be run as a service and as a console
- [x] consumes messages (for now of type `JsonMetricsContext`)
- [x] dispatches them to an asynchronous consumer (ConcurrentQueue)
- [x] enables registering consumers
- [ ] gathers raw metrics


Scooletz:
Ping @SzymonPobiega 
Could you review it?

src/ServiceControl.Monitoring/MetricReport.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace NServiceBus.Metrics
+{
+    using Newtonsoft.Json.Linq;
+
+    /// <summary>
+    /// The reporting message.
+    /// </summary>
+    public class MetricReport : ICommand
```
weralabaj: That could simply by `IMessage`.

SzymonPobiega: I think we should de-serialize directly to JsonMetricsContext instead of JObject. It should be possible if we use Newtonsoft serializer as daniel pointed out.

src/ServiceControl.Monitoring/MetricsHandler.cs
```diff
@@ -1,20 +1,20 @@
...
 
-    public class MetricsHandler : IHandleMessages<JsonMetricsContext>
+    public class MetricsHandler : IHandleMessages<MetricReport>
     {
         static readonly Task<int> CompletedTask = Task.FromResult(0);
-        readonly PublisherConsumer<JsonMetricsContext> publisherConsumer;
+        readonly PublisherConsumer<MetricReport> publisherConsumer;
 
-        public MetricsHandler(PublisherConsumer<JsonMetricsContext> publisherConsumer)
+        public MetricsHandler(PublisherConsumer<MetricReport> publisherConsumer)
```
SzymonPobiega: Maybe out of scope of this PR but why going through `PublisherConsumer` instead of injecting `NServiceBusReceivedMetricContext` to the handler via container?

weralabaj:
LGTM

src/ServiceControl.Monitoring/MetricReport.cs
weralabaj: So we assume here that for now we ignore the potential extensions, but potentially will use it in the future?

SzymonPobiega: Extensions?

weralabaj: Probably "customizations" would be a better word. I mean that even though officially that could be anything, we still make an assumption this will be the exact type used, and if NSB.Metrics adds/modifies some data, then we'll break.

src/ServiceControl.Monitoring/MetricsHandler.cs
Scooletz: I was planning to get `NServiceBusReceivedMetricContext` registered as a consumer as every other tool.

src/ServiceControl.Monitoring/MetricReport.cs
Scooletz: After taking a look how the raw data were reported with `NServiceBusReceivedMetricContext` I think that I agree with @weralabaj . The reporting could be done as a [TolerantReader](https://martinfowler.com/bliki/TolerantReader.html) that simply extracts `JProperty` with name `Context` and store the whole JObject payload in a dictionary. This would remove our dependency on the Metrics leaving this data as raw as possible.

SzymonPobiega: I am not sure what you are trying to achieve here. NServiceBus.Metrics depends on Metrics.NET and uses it to generate Json. We will use Metrics.NET here internally to generate derived metrics (most likely). What would we gain from pretending we don't use Metrics.NET?

src/ServiceControl.Monitoring/MetricsHandler.cs
SzymonPobiega: But we don't expect any other consumers, do we?

Scooletz:
Recent commits' changes:
- code aligned with the new format of the message from the metrics plugin
- raw data are reported with raw `JObject` JSON structures
- `Metrics.NET` removed as they are not needed for now.

Scooletz:
Ping @weralabaj @SzymonPobiega

weralabaj:
LGTM. As for the type, I think until proven it's impossible/too much effort, we should not rely on Metrics.NET types. ATM it seems it'll be simpler that way.

SzymonPobiega:
:+1:. Let's get this in.

PR #3, opened 2017-04-07
Move target import

danielmarbach:


Scooletz:
It looks that this PR did not change the build status. Also, after tagging develop we're green so far.

PR #9, opened 2017-04-19
Http endpoint configuration exposed in App.config

Scooletz:
This PR currently does not include tests for configuration keys. Any ideas how to test it ?

### DIFF
--- a/src/ServiceControl.Monitoring/Http/HttpEndpoint.cs
+++ b/src/ServiceControl.Monitoring/Http/HttpEndpoint.cs
@@ -8,8 +8,15 @@ using NServiceBus.ObjectBuilder;
 
 namespace ServiceControl.Monitoring.Http
 {
+    using System;
+    using System.Configuration;
+    using System.Linq;
+
     class HttpEndpoint : Feature
     {
+        public const string AppConfigHostnameKey = "HttpEndpoint/Hostname";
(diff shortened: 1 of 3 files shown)


src/ServiceControl.Monitoring/Http/HttpEndpoint.cs
```diff
@@ -8,31 +8,55 @@
...
+            var hostname = GetValue(AppConfigHostnameKey);
+            var port = GetValue(AppConfigPort);
+            int portValue;
+
+            if (string.IsNullOrWhiteSpace(hostname) == false &&
+                string.IsNullOrWhiteSpace(port) == false &&
+                int.TryParse(port, out portValue))
+            {
+                host = new Uri(host + ":" + port);
+            }
```
SzymonPobiega: should we do something if `TryParse` fails? I mean, if somebody put `dupa` in the port number e.g. `1234dupa`?

Scooletz: Currently it falls back to the default value. should it throw?

SzymonPobiega: I think it should. I mean absence of value is fine but non-integer should throw.

PR #10, opened 2017-04-26
Introduce queue-length metric

Scooletz:
Connects to https://github.com/Particular/LaunchWave.DevOpsAdvancedMonitoring/issues/66

This PR introduces the queue-length metric with following steps:
- a new extension point is provided for metrics that are reported via http. It's called `IEndpointDataProvider` 
- every `IEndpointDataProvider` is reported under its own JProperty (see below)
- `RawDataProvider` is now `IEndpointDataProvider`
- queue-length is built on the basis of messages sent by `NServiceBus.Metrics` reporting sequence numbers for a specified session for a specified queue
- queue-length is reported alongside to the raw metrics as it's grouped by queue (which effectively determines the endpoint)

Schema for http report accessible under `/metrics/raw`:

```json
{
  "NServiceBus.Endpoints" : {
    "Endpoint1" : {
      "Raw" : {},
      "QueueLength" : 100
    },
    "Endpoint2" : {
      "Raw" : {},
      "QueueLength" : 2300
    }
  }
}
```


Scooletz:
Ping @SzymonPobiega @seanfarmar for a review of the new approach for reporting metrics.

SzymonPobiega:
LGMT so far. @Scooletz let me know when you want it merged.

Scooletz:
As soon as I got the metric implemented. Should be today. I'll ping you again.

Scooletz:
I've update this PR with the receive part. Currently, the test is failing as it looks like another session is somehow generated. Working on it.

src/ServiceControl.Monitoring/QueueLength/QueueLengthCalculator.cs
```diff
@@ -0,0 +1,101 @@
...
+            var receives = gauges.Where(c => c.Value<string>("Name").StartsWith("QueueLengthReceive_"));
+            UpdateReceives(receives, endpoint);
+        }
+
+        void UpdateSends(IEnumerable<JToken> sends)
+        {
+            foreach (var send in sends)
+            {
+                var name = send.Value<string>("Name");
+                var sessionId = Guid.Parse(name.Substring(name.LastIndexOf("_") + 1));
```
SzymonPobiega: Can sessionID be a tag on the metric instead of being part of the name?

Scooletz: I'd say, let's stick to the semantics of tags: https://github.com/etishor/Metrics.NET/wiki/Grouping-&-Organizing-Metrics#tags and put the "QueueLengthSend" or "Receive" in there, keeping the session id in the name of the counter/gauge.

SzymonPobiega: How about the destination or message type? Would that be part of the name or also a tag?

src/ServiceControl.Monitoring.AcceptanceTests/Tests/Endpoint_using_metrics_provider.cs
```diff
@@ -16,39 +16,74 @@
...
+                            }
+                        }
                     }
                 }
             };
 
             Assert.AreEqual(expected.ToString(Formatting.None), context.CrawlResult);
         }
 
+        //[Test]
```
Scooletz: Why is this test commented?

src/ServiceControl.Monitoring.Tests/QueueLength/CalculatorTests.cs
```diff
@@ -0,0 +1,80 @@
+namespace ServiceControl.Monitoring.Tests.QueueLength
+{
+    using Monitoring.QueueLength;
+    using NUnit.Framework;
+
+    public class CalculatorTests
+    {
+        [Test]
+        public void When_only_receiver_reports_Count_should_be_zero()
```
Scooletz: A nice approach with extracting the calculator

src/ServiceControl.Monitoring.Tests/QueueLength/QueueLengthDataConsumerTests.cs
```diff
@@ -0,0 +1,144 @@
...
+    using System.Collections.Generic;
+    using Monitoring.QueueLength;
+    using Newtonsoft.Json.Linq;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class QueueLengthDataConsumerTests
+    {
+        [Test]
+        public void It_extracts_sent_sequence()
```
Scooletz: Why to we put JSON in here instead of serializing data using `Metrics` API? To ensure compatibility when the package is updated? IMO this test is fragile.

src/ServiceControl.Monitoring/QueueLength/QueueLengthCalculator.cs
```diff
@@ -0,0 +1,55 @@
...
+    using System;
+    using System.Collections.Concurrent;
+    using System.Collections.Generic;
+
+    class QueueLengthCalculator : IQueueLengthCalculator
+    {
+        public void UpdateReceivedSequence(string key, long value, string queue)
+        {
+            receivedSequences.AddOrUpdate(key, _ => value, (_, prev) => Math.Max(prev, value));
+            sessionKeyToQueueMapping.AddOrUpdate(key, queue, (_, __) => queue);
```
Scooletz: `.GetOrAdd` is cheaper and aquires no lock for existing values.

src/ServiceControl.Monitoring/QueueLength/QueueLengthCalculator.cs
```diff
@@ -0,0 +1,55 @@
...
+        {
+            var queueLengths = new Dictionary<string, long>();
+
+            foreach (var kvp in sessionKeyToQueueMapping)
+            {
+                var queueName = kvp.Value;
+
+                if (!queueLengths.ContainsKey(queueName))
+                {
+                    queueLengths[queueName] = 0;
```
Scooletz: extract `queueLengths[queueName]` to a variable, if possible?

src/ServiceControl.Monitoring/QueueLength/TagExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+        {
+            throw new Exception($"Tag {key} not found.");
+        }
+        return result;
+    }
+
+    public static bool TryGetTagValue(this IEnumerable<string> tags, string key, out string value)
+    {
+        value = tags
+            .Where(t => t.StartsWith($"{key}:", StringComparison.OrdinalIgnoreCase))
```
Scooletz: extract interpolation to a variable

Scooletz: Should we consider writing functions for all the key values? Like

`TryGetTypeTagValue` and have strings consts?

src/ServiceControl.Monitoring/ServiceControl.Monitoring.csproj
```diff
@@ -12,11 +12,26 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>ServiceControl.Monitoring</RootNamespace>
     <AssemblyName>ServiceControl.Monitoring</AssemblyName>
-    <TargetFrameworkVersion>v4.5.2</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.6.2</TargetFrameworkVersion>
```
Scooletz: Why do we target `4.6.2`?

src/ServiceControl.Monitoring.Tests/QueueLength/QueueLengthDataConsumerTests.cs
SzymonPobiega: Because we said we want to defer the dependency on Metrics here as long as we can. This json is the wire format we accept. Plus it would be quite cumbersome to get that json via metrics and it would not be evident what values are there.

Scooletz: `This json is the wire format we accept.` agreed.

src/ServiceControl.Monitoring/QueueLength/QueueLengthCalculator.cs
SzymonPobiega: I've refactored it a bit. Take a look (after I push;-)

src/ServiceControl.Monitoring/QueueLength/TagExtensions.cs
SzymonPobiega: Nah, I don't think it is worth the effort.

src/ServiceControl.Monitoring/ServiceControl.Monitoring.csproj
SzymonPobiega: What should we target? ;-)

timbussmann: core currently targets 4.5.2 and will move to 4.6.1 or higher with the next major.

SzymonPobiega:
@Scooletz I fixed the acceptance tests.

src/ServiceControl.Monitoring/ServiceControl.Monitoring.csproj
SzymonPobiega: Well, this is a stand-alone service so I guess we can chose whatever we want. What value makes sense these days?

timbussmann: depends on whether you need to use some of the 4.6.x changes or not. Otherwise it would be more appropriate to use 4.5.2 as the core so you won't force users to upgrade just because they want to use this plugin?

SzymonPobiega: It is not a plugin. It is a service, similar to SC.

gbiellem: If this targets anything higher that 4.5.2 then we need to update the PI and the MSI for ServiceControl.  The PI installs 4.5.2 if it is not present and the SC MSI install has checks in it to determine whether the required framework is there

SzymonPobiega:
@timbussmann @Scooletz I switched back to 452.

PR #11, opened 2017-05-02
Packaging for SCMU

gbiellem:
Related to https://github.com/Particular/LaunchWave.DevOpsAdvancedMonitoring/issues/51

- [x] Package up v6 transports and service for SCMU (zip)
- [x] Add app.config settings for transport choice and connection string
- [x] Add app.config settings for endpoint name
- [x] Add app.config settings and logic for log path
- [x] Add command line options to run installers and exit
- [x] Disable installers being enabled by default
- [x] Implement Custom ReleaseDate attinbute for licensing checks
- [x] Modify Nuget output to contain zip so we can ref in SCMU
- [x] Add Octopus deployment to MyGet


boblangley:
Adding log path / level settings and made logging more like SC logging

boblangley:
I checked all the boxes. As far as I can tell we did all those things.

gbiellem:
I added two more tasks, both are pretty trivial, I started on the nuget packaging.  I plan to put the zip and an ips1 file into the nuget so it copies the zip to the correct place in the SC project.

We'll also need to adjust deployment as this nuget should only reside on myget since it's only a delivery mecahnism for the zip

gbiellem:
I've added octopus deployment steps added to TC and Octopus so we can push to Myget once this is ready for merged.

gbiellem:
I think all yak shaving is complete. Merge?

gbiellem:
Bump @boblangley  @mikeminutillo ☝️

src/ServiceControl.Monitoring/Settings.cs
```diff
@@ -0,0 +1,37 @@
...
+        public string Username { get; set; }
+        public bool EnableInstallers { get; set; }
+
+        internal static Settings Load(SettingsReader reader)
+        {
+            var settings = new Settings
+            {
+                TransportType = reader.Read("Transport/Type", typeof(MsmqTransport).AssemblyQualifiedName),
+                TransportConnectionString = reader.Read<string>("Transport/ConnectionString"),
+                LogPath = CalculateLogPathForEndpointName(DEFAULT_ENDPOINT_NAME),
```
gbiellem: @boblangley any reason this isn't configurable? Would be nicer to allow them to override the default same as SC

gbiellem:
My OCD is kicking in...

The keys in the app config setting naming doesn't align with SC. 
There the keys use the app name . i.e. ""ServiceControl/TransportType"  but here the key is based on the item being configured so here we have ""Transport/Type" and  "HttpEndpoint/Hostname". Are we happy with that or should the two be consistent?

Also connections strings in ServiceControl are not stored in appSettings but are actually in the connection string section as a named connection string called "NServiceBus/Transport".  I think this should also be aligned

mikeminutillo:
@gbiellem what should the keys be? `ServiceControl/TransportType` or `ServiceControl.Monitoring/TransportType`? I named them that way when I was considering introducing an [EasySettings](http://codermike.com/convention-config-for-autofac) Autofac module. That would have allowed us to just have:

```csharp
class TransportSettings
{
  public string Type { get; set; } = typeof(MsmqTransport).AssemblyQualifiedName;
  public string ConnectionString { get; set; }
}
```

It makes adding new settings extremely easy. I didn't end up doing it because it didn't seem like we had enough settings to make it worth it, and we generally need settings before setting up IoC.

gbiellem:
> @gbiellem what should the keys be? `ServiceControl/TransportType` or `ServiceControl.Monitoring/TransportType`?

I'd lean towards prefixing with Monitoring i.e.  `Monitoring/TransportType`. The full name seems too long,

boblangley:
👍 to `Monitoring/*'

mikeminutillo:
OK. I'm done now. If the build is green I'm happy. Do we need more testing around the configurable bits? Do we need to test on multiple transports?

src/ServiceControl.Monitoring/Settings.cs
boblangley: Just to point out, it was set to override based on the endpoint name when passed in at the command line. Since your comment (which I missed, apologies) Mike refactor it to call the method with the endpoint name when the LogPath property is accessed.

boblangley:
@mikeminutillo @gbiellem As far as I can tell we don't test these things with Service Control yet. 

Not to say that testing that configuring service name, log paths, transports, etc isn't important, but I think we should have those covered under another task force. Possibly even pushed off onto the maintainer group.

With that in mind I am going to merge this PR. We can open another to add the testing you described if you disagree.

PR #19, opened 2017-06-02
Starting endpoint instance in setup command

### LINKED ISSUE #17: When running only -s (setup command) the installers are not run
When running -s command the result is that [`Endpoint.Create`](https://github.com/Particular/ServiceControl.Monitoring/blob/develop/src/ServiceControl.Monitoring/Hosting/Commands/SetupCommand.cs#L12) is called which does not run installers. We probably need to call `Endpoint.Start` to make that happen.

tmasternak:
Resolves #17


mikeminutillo:
`Endpoint.Create` already runs the installers.

PR #21, opened 2017-06-08
Timing Data Aggregation and API for SP Endpoint Overview Page

WojcikMike:
Connects to Particular/LaunchWave.DevOpsAdvancedMonitoring#52

### Description
This PR is adding storage of historical data (last 5 min) for ProcessingTime and CriticalTime metrics. In particular it introduces
 * Deserializer for `LongValueOccurence` content type 
 * API module to enable querying for that data `DiagramsApiModule`
 * For each logical endpoint it stores metrics data in 15s intervals
 * It removes `IDataProvider`, `IDataConsumer` interfaces as those can replaced with multiple NSB handlers for the same message type e.g. `MetricsReport`


src/ServiceControl.Monitoring/Diagrams/DiagramDataProvider.cs
```diff
@@ -0,0 +1,52 @@
...
+    public class DiagramDataProvider : IRawDataConsumer
+    {
+        public MonitoringData MonitoringData { get; } = new MonitoringData(10);
+
+        /// <summary>
+        /// Consumes a new portion of data.
+        /// </summary>
+        public void Consume(IReadOnlyDictionary<string, string> headers, JObject data)
+        {
+            JsonConvert.SerializeObject(data["Timestamp"]);
```
Scooletz: Why is this call needed? The value is obtained below.

src/ServiceControl.Monitoring/Diagrams/MonitoringData.cs
```diff
@@ -0,0 +1,55 @@
...
+
+            this.size = size;
+            head = 0;
+        }
+
+        public void Record(DateTime timestamp, float? criticalTime, float? processingTime)
+        {
+            Interlocked.Increment(ref head);
+
+            var index = (head - 1) % size;
```
Scooletz: Use the value returned from `Interlocked.Increment` rather than rereading the field (which depending on various conditions might return a different value)

src/ServiceControl.Monitoring/Diagrams/MonitoringData.cs
```diff
@@ -0,0 +1,55 @@
...
+
+            Timestamps[index] = timestamp;
+            CriticalTime[index] = criticalTime;
+            ProcessingTime[index] = processingTime;
```
Scooletz: All the values are written with no memory fence. Effectively their assignments might be postponed in time as the reader sets no barrier as well. Instead of having a class with 3 arrays, we could think about having an array of structs with three fields `Timestamp`, `CriticalTime`, `ProcessingTime` and one additional Status for a barrier. When writing under specific index, you could assign all the values and `Volatile.Write (ref records[index].Barrier, 1)` the barrier.
When reading, first the `Volatile.Read (ref records[index].Barrier)` would be called with followed reads of the rest of items.
It might be not needed if the visibility of these values is ensured in another way (but I see nothing like this).

src/ServiceControl.Monitoring/Http/HttpEndpoint.cs
```diff
@@ -52,7 +53,7 @@ class NancyTask : FeatureStartupTask
 
             public NancyTask(IBuilder builder, Uri host)
             {
-                var buildstrapper = new Bootstrapper(builder.BuildAll<IEndpointDataProvider>());
+                var buildstrapper = new Bootstrapper(builder.BuildAll<IEndpointDataProvider>(), builder.Build<DiagramDataProvider>());
```
Scooletz: Maybe we should invest in making it impl-agnostic and based only on the all data providers? With this approach, there will be an additional registration per provider.

src/ServiceControl.Monitoring.Tests/Timings/TimingsStoreTests.cs
```diff
@@ -0,0 +1,156 @@
...
+
+            store.Store(string.Empty, message, now);
+
+            var timings = store.GetTimings(now);
+
+            Assert.IsTrue(timings[0].Intervals.All(i => i.Average == 0));
+        }
+
+        [Test]
+        public void Intervals_from_the_future_are__stored()
```
SzymonPobiega: Double `_`

src/ServiceControl.Monitoring/EndpointFactory.cs
```diff
@@ -39,12 +41,14 @@ public static void MakeMetricsReceiver(EndpointConfiguration config, Settings se
             config.GetSettings().Set<Settings>(settings);
 
             config.UseSerialization<NewtonsoftSerializer>();
+            config.AddDeserializer<LongValueOccurrenceSerializerDefinition>();
             config.UsePersistence<InMemoryPersistence>();
             config.SendFailedMessagesTo(settings.ErrorQueue);
             config.LimitMessageProcessingConcurrencyTo(1);
```
SzymonPobiega: Do we still need to enforce concurrency=1 ?

src/ServiceControl.Monitoring/Metrics/Raw/LongValueOccurenceSerializerDefinition.cs
```diff
@@ -0,0 +1,59 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+    using System.IO;
+    using global::NServiceBus.MessageInterfaces;
+    using global::NServiceBus.Serialization;
+    using global::NServiceBus.Settings;
+    using NServiceBus.Metrics;
+
+    class LongValueOccurrenceSerializerDefinition : SerializationDefinition
```
SzymonPobiega: Using a serializer requires execution of full pipeline. I think that for the purpose of these messages we would be better off using a behavior in the transport receive pipeline. This way we would shortcut the pipeline saving us some processing cycles.

The behavior would replace both serializer and the handler.

tmasternak: I agree that performance is not enough with current approach. That being said the structure of data stored in memory will very soon change to include longer historical intervals and data for purposes of details page. 
As a result the plan is to wait until we have better knowledge about what we want to store and tackle performance at that point.

src/ServiceControl.Monitoring/EndpointFactory.cs
tmasternak: removed

src/ServiceControl.Monitoring.AcceptanceTests/ServiceControl.Monitoring.AcceptanceTests.csproj
```diff
@@ -42,6 +42,9 @@
       <HintPath>..\packages\NServiceBus.6.2.0\lib\net452\NServiceBus.Core.dll</HintPath>
       <Private>True</Private>
     </Reference>
+    <Reference Include="NServiceBus.Metrics, Version=1.0.0.0, Culture=neutral, PublicKeyToken=9fc386479f8a226c, processorArchitecture=MSIL">
```
timbussmann: where is this used?

src/ServiceControl.Monitoring/Timings/TimingsStore.cs
```diff
@@ -0,0 +1,154 @@
...
+                    }
+                }
+
+            }
+        }
+
+        public EndpointTimings[] GetTimings(DateTime now)
+        {
+            var result = new List<EndpointTimings>();
+            var intervals = GenerateIntervalIds(now, NumberOfHistoricalIntervals);
```
mikeminutillo: Here you need this because you want to include blank periods in the result right?

src/ServiceControl.Monitoring/Timings/TimingsStore.cs
```diff
@@ -0,0 +1,154 @@
...
+                    };
+
+                    result.Add(endpointTimings);
+                }
+            }
+
+            return result.ToArray();
+        }
+
+        static DateTime[] GenerateIntervalIds(DateTime now, int numberOfPastIntervals)
```
mikeminutillo: Should this be `now` or `start`?

src/ServiceControl.Monitoring/Timings/TimingsStore.cs
```diff
@@ -0,0 +1,154 @@
...
+        static DateTime IntervalId(DateTime date)
+        {
+            var interval = 0;
+
+            while (interval + IntervalSizeInSec < date.Second)
+            {
+                interval += IntervalSizeInSec;
+            }
+
+            return new DateTime(date.Year, date.Month, date.Day, date.Hour, date.Minute, interval, DateTimeKind.Utc);
```
mikeminutillo: If it's just rounding down to the nearest interval value then you can do this:

```
public static DateTime RoundDownToNearest(this DateTime dateTime, TimeSpan timespan)
	=> new DateTime(dateTime.Ticks - dateTime.Ticks % timespan.Ticks, dateTime.Kind);
```

Then you can just call `now.RoundDownToLatest(TimeSpan.FromSeconds(15))`

src/ServiceControl.Monitoring/Http/QueueLengthApiModule.cs
```diff
@@ -0,0 +1,44 @@
...
+    using Newtonsoft.Json;
+    using Newtonsoft.Json.Linq;
+    using QueueLength;
+
+    public class QueueLengthApiModule : NancyModule
+    {
+        /// <summary>
+        /// Initializes the metric API module.
+        /// </summary>
+        public QueueLengthApiModule(QueueLengthDataStore store) : base("/metrics")
```
tmasternak: This is a result of refactoring the bits that were already there. I propose to leave this as is and make proper changes when we start delivering features around `QueueLength`.

src/ServiceControl.Monitoring/TaskEx.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace ServiceControl.Monitoring
+{
+    using System.Threading.Tasks;
+
+    public class TaskEx
+    {
+        public static Task Completed = Task.FromResult(0);
```
tmasternak: I don't have a good reason. Calling @Particular/servicecontrol-maintainers 🦇

src/ServiceControl.Monitoring.AcceptanceTests/ServiceControl.Monitoring.AcceptanceTests.csproj
tmasternak: In [when_querying_timing_data](https://github.com/Particular/ServiceControl.Monitoring/pull/21/files#diff-46cce0a428a906b09e88829defd86bcfR53) to setup metric reporting in the monitored endpoint.

src/ServiceControl.Monitoring/Http/DiagramApiModule.cs
```diff
@@ -0,0 +1,65 @@
...
+                var criticalTimes = criticalTimeStore.GetTimings(now);
+
+                var endpointNames = processingTimes.Select(e => e.EndpointName)
+                    .Union(criticalTimes.Select(e => e.EndpointName));
+
+                var endpointsResult = new List<JObject>();
+
+                foreach (var endpointName in endpointNames)
+                {
+                    var processingTime = processingTimes.FirstOrDefault(pt => pt.EndpointName == endpointName);
```
tmasternak: I propose to leave out the performance concerns for now. See previous [comment](https://github.com/Particular/ServiceControl.Monitoring/pull/21#discussion_r126403236). 

In short I think that digging into performance now will be a wasted effort.

src/ServiceControl.Monitoring/Timings/LongValueOccurrenceHandler.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        public TimingsReportHandler(ProcessingTimeStore processingTimeStore, CriticalTimeStore criticalTimeStore)
+        {
+            this.processingTimeStore = processingTimeStore;
+            this.criticalTimeStore = criticalTimeStore;
+        }
+
+        public Task Handle(LongValueOccurrences message, IMessageHandlerContext context)
+        {
+            var messageType = context.MessageHeaders[Headers.EnclosedMessageTypes];
```
mikeminutillo: Should we use a different header for these? Or should it be a part of the payload to send the vector type in? If someone changes the inheritance hierarchy of this message type, the code will break (silently)

src/ServiceControl.Monitoring/Timings/LongValueOccurrenceHandler.cs
```diff
@@ -0,0 +1,40 @@
...
+            var endpointName = context.MessageHeaders.GetOriginatingEndpoint();
+
+            if (messageType == ProcessingTimeMessageType)
```
mikeminutillo: Should this be a switch? What if we get a message type we don't understand?

src/ServiceControl.Monitoring/Http/DiagramApiModule.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// Initializes the metric API module.
+        /// </summary>
+        public DiagramsApiModule(ProcessingTimeStore processingTimeStore, CriticalTimeStore criticalTimeStore) : base("/diagrams")
+        {
+            After.AddItemToEndOfPipeline(ctx => ctx.Response
+                .WithHeader("Access-Control-Allow-Origin", "*")
+                .WithHeader("Access-Control-Allow-Methods", "POST,GET")
+                .WithHeader("Access-Control-Allow-Headers", "Accept, Origin, Content-type"));
+
+            Get["/data"] = x =>
```
tmasternak: `/diagrams` it is

src/ServiceControl.Monitoring/Http/DiagramApiModule.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// <summary>
+        /// Initializes the metric API module.
+        /// </summary>
+        public DiagramsApiModule(ProcessingTimeStore processingTimeStore, CriticalTimeStore criticalTimeStore) : base("/diagrams")
+        {
+            After.AddItemToEndOfPipeline(ctx => ctx.Response
+                .WithHeader("Access-Control-Allow-Origin", "*")
+                .WithHeader("Access-Control-Allow-Methods", "POST,GET")
+                .WithHeader("Access-Control-Allow-Headers", "Accept, Origin, Content-type"));
+
```
mikeminutillo: Should this stuff be moved to proper middleware (or a base module class?

src/ServiceControl.Monitoring/Http/HttpEndpoint.cs
```diff
@@ -52,9 +52,13 @@ class NancyTask : FeatureStartupTask
 
             public NancyTask(IBuilder builder, Uri host)
             {
-                var buildstrapper = new Bootstrapper(builder.BuildAll<IEndpointDataProvider>());
+                var bootstrapper = new Bootstrapper(
+                    builder.Build<QueueLengthDataStore>(),
```
tmasternak: I've removed the dependency from `IBuilder` and now we are passing actions that do concrete type registrations.

src/ServiceControl.Monitoring/Http/DiagramApiModule.cs
```diff
@@ -0,0 +1,65 @@
...
+                                { "points", new JArray(criticalTime?.Intervals.OrderBy(i => i.IntervalStart).Select(i => i.Average)) }
+                            }
+                        }
+                    });
+                }
+
+                var result = new JArray(endpointsResult);
+
+                return Response.AsText(result.ToString(Formatting.None), "application/json");
+            };
```
mikeminutillo: Why are we doing low-level Json here instead of having an object graph and allowing Nancy to serialize?

src/ServiceControl.Monitoring/Metrics/Raw/LongValueOccurrences.cs
```diff
@@ -0,0 +1,28 @@
+﻿namespace ServiceControl.Monitoring.Processing.RawData.NServiceBus.Metrics
```
tmasternak: LOL. Fixed

src/ServiceControl.Monitoring/QueueLength/QueueLengthFeature.cs
```diff
@@ -7,7 +7,7 @@ class QueueLengthFeature : Feature
     {
         protected override void Setup(FeatureConfigurationContext context)
         {
-            context.Container.ConfigureComponent(b => new QueueLengthDataConsumer(new QueueLengthCalculator()), DependencyLifecycle.SingleInstance);
+            context.Container.ConfigureComponent(b => new QueueLengthDataStore(new QueueLengthCalculator()), DependencyLifecycle.SingleInstance);
```
tmasternak: done

src/ServiceControl.Monitoring/Timings/TimingsStore.cs
tmasternak: Once again I would propose to hold-off. I think that what is missing here is a value type expressing concept of an interval. In such case the hashing algorithm would hash over intervals.

Right now the whole storage is encapsulated in a single class so I would not worry about dangerous details leaking out.

src/ServiceControl.Monitoring/Http/DiagramApiModule.cs
mikeminutillo: Would it make sense to have 3 different calls here? Get Endpoints, Get Processing Time, Get Critical Time? That will change the UI from 2 remote calls to 5 but it gives us greater flexibility.

src/ServiceControl.Monitoring/QueueLength/QueueLengthDataStore.cs
```diff
@@ -2,13 +2,11 @@ namespace ServiceControl.Monitoring.QueueLength
 {
     using System.Collections.Generic;
     using System.Linq;
-    using Http;
     using Newtonsoft.Json.Linq;
-    using Raw;
 
-    class QueueLengthDataConsumer : IRawDataConsumer, IEndpointDataProvider
+    public class QueueLengthDataStore
```
mikeminutillo: This "store" and the others (critical/processing time) seem to operate on different levels of abstraction. This one wants json and extracts the info it needs. The others want values already extracted from the message. I'm not sure it really matters but I thought I'd point it out.

src/ServiceControl.Monitoring/Timings/TimingsStore.cs
tmasternak: Correct.

src/ServiceControl.Monitoring/Http/DiagramApiModule.cs
tmasternak: I think it's a matter of asking if we are building API or Remote Interface. 

I think that the current assumption was to build Remote Interface, driven only by SP requirements and as a result have only a single call.

src/ServiceControl.Monitoring/Timings/TimingsStore.cs
```diff
@@ -0,0 +1,154 @@
...
+    using Processing.RawData.NServiceBus.Metrics;
+
+    public abstract class TimingsStore
+    {
+        ConcurrentDictionary<string, ConcurrentDictionary<DateTime, MeasurementInterval>> timings = 
+            new ConcurrentDictionary<string, ConcurrentDictionary<DateTime, MeasurementInterval>>();
+
+        public void Store(string endpointName, LongValueOccurrences message, DateTime now)
+        {
+            var endpointData = timings.GetOrAdd(endpointName, new ConcurrentDictionary<DateTime, MeasurementInterval>());
```
tmasternak: Hmm I think I'm not following. Will it? According to [msdn](https://msdn.microsoft.com/en-us/library/ee378676(v=vs.110).aspx) it will create new instance if none exists. (I'm probably missing something here).

src/ServiceControl.Monitoring/Timings/LongValueOccurrenceHandler.cs
tmasternak: Done.

src/ServiceControl.Monitoring/Timings/TimingsStore.cs
```diff
@@ -0,0 +1,154 @@
...
+                foreach (var interval in endpointData.Keys)
+                {
+                    if (validIntervals.Contains(interval) == false)
+                    {
+                        MeasurementInterval bucket;
+                        endpointData.TryRemove(interval, out bucket);
+                    }
+                }
+
+            }
```
tmasternak: The data is stored only for non-empty intervals so just counting the intervals will not do the trick I'm afraid. 
When it comes to passing `now` it also enables writing predictable [tests](https://github.com/Particular/ServiceControl.Monitoring/pull/21/files#diff-8c25298ee0255593d53dc1673450ce55R11).

tmasternak: That being said I changed this a bit not to generate intervals but to calculate oldest valid interval and remove anything else.

src/ServiceControl.Monitoring/Timings/LongValueOccurrenceHandler.cs
tmasternak: On the sending side there is no `LongValueOccurence` message at all. There is only a raw dispatcher that writes the bytes. The deserialization relies on two headers right now `ContentType` and `EclosedMessagesTypes`. First to identify the metrics message format and second to identify the name of the metric it holds.

src/ServiceControl.Monitoring/Http/DiagramApiModule.cs
tmasternak: done

src/ServiceControl.Monitoring/Http/DiagramApiModule.cs
tmasternak: Introduced proper DTO.

src/ServiceControl.Monitoring/QueueLength/QueueLengthDataStore.cs
tmasternak: I agree that the naming is far from ideal. I would propose to wait with name changes until we do https://github.com/Particular/LaunchWave.DevOpsAdvancedMonitoring/issues/109 which has just started.

tmasternak:
@mikeminutillo I've addressed your review comments. Can you have a look?

src/ServiceControl.Monitoring/Timings/TimingsStore.cs
mikeminutillo: The MSIL will look something like
```
LOAD endpointname
LOAD new ConcurrentDictionary
CALL GetOrAdd
```
It needs to call the constructor in order to pass the object in to the `GetOrAdd` call. There should be an overload of `GetOrAdd` that takes a `Func` that can delay that.

src/ServiceControl.Monitoring/Timings/LongValueOccurrenceHandler.cs
mikeminutillo: `EnclosedMessageTypes` is already an NServiceBus header and we're kind of using it differently. Would it be OK to call this `MetricTypes`?

tmasternak:
@mikeminutillo back to you :)

src/ServiceControl.Monitoring/TaskEx.cs
mikeminutillo: No particular reason. Feel free to bump up

src/ServiceControl.Monitoring/Timings/LongValueOccurrenceHandler.cs
```diff
@@ -0,0 +1,50 @@
...
+                    processingTimeStore.Store(endpointName, message, DateTime.UtcNow);
+                    break;
+                case CriticalTimeMessageType:
+                    criticalTimeStore.Store(endpointName, message, DateTime.UtcNow);
+                    break;
+                default:
+                    throw new UnknownLongValueOccurrenceMessageType(messageType);
+
+            }
+            if (messageType == ProcessingTimeMessageType)
```
mikeminutillo: I presume this can just be deleted

tmasternak:
> I wonder if we can keep data points in a circular buffer (for each endpoint) rather than dictionaries that we need to constantly filter, but we can address that later because it's an implementation detail.

💯 👍

Scooletz:
@mikeminutillo @tmasternak 
A clever data structure could be important in here. We want to measure more than 10 endpoints, right? ;-)

src/ServiceControl.Monitoring/Timings/TimingsStore.cs
```diff
@@ -0,0 +1,141 @@
...
+    using Metrics.Raw;
+
+    public abstract class TimingsStore
+    {
+        ConcurrentDictionary<string, ConcurrentDictionary<DateTime, MeasurementInterval>> timings = 
+            new ConcurrentDictionary<string, ConcurrentDictionary<DateTime, MeasurementInterval>>();
+
+        public void Store(string endpointName, LongValueOccurrences message, DateTime now)
+        {
+            var endpointData = timings.GetOrAdd(endpointName, _ => new ConcurrentDictionary<DateTime, MeasurementInterval>());
```
Scooletz: It's good to have the data partitioned by `endpoint` as this is the way we want to display/analyze them.

src/ServiceControl.Monitoring/Metrics/Raw/LongValueOccurenceSerializerDefinition.cs
```diff
@@ -0,0 +1,58 @@
...
+            {
+                message
+            };
+        }
+
+        static void ToLongValueOccurrences(LongValueOccurrences message, Stream stream)
+        {
+            var reader = new BinaryReader(stream);
+
+            message.Version = reader.ReadInt64();
```
Scooletz: Allocating arrays for ticks and values could lead to a lot of overhead. If we want to use `IMessageSerializer` we could create a pool of `LongValueOccurences` that enables leasing messages and passing them further with no allocations.
An interface could be like:

```
interface IPool 
{
   LongValueOccurrences  Lease (int minCountOfEntries);
   void Return (LongValueOccurences messagE);
}
```

with `Lease` used in the deserialization and `Release` at the end of the handler.

src/ServiceControl.Monitoring/Timings/TimingsStore.cs
```diff
@@ -0,0 +1,141 @@
...
+
+            for (var i = 0; i < message.Ticks.Length; i++)
+            {
+                var date = new DateTime(message.BaseTicks + message.Ticks[i], DateTimeKind.Utc);
```
Scooletz: If instead of deserializing data to a `LongValueOccurrences` and process it in here, we introduce a simple array of measurements being just an array of {Time, Value} measurements, we could do this work in the deserialization. Moreover, we could use pooling for these arrays and simply add this to a list of measurements for specific `endpointData`. This would lead to tranforming `endpointData` from a concurrent dictionary to a simple linked list of chunks of measurements (one chunk per incoming message).

src/ServiceControl.Monitoring/Timings/TimingsStore.cs
```diff
@@ -0,0 +1,141 @@
...
+                var intervalId = date.RoundDownToNearest(IntervalSize);
+
+                endpointData.AddOrUpdate(
+                    intervalId,
+                    _ => new MeasurementInterval(1, message.Values[i]),
+                    (_, b) => b.Update(message.Values[i]));
+            }
+
+            if (endpointData.Count > 2 * NumberOfHistoricalIntervals)
```
Scooletz: With the proposal above, cleaning oldest interval is just walking through the list and `Release`ing as many chunks as needed, starting from the head.

src/ServiceControl.Monitoring/Timings/TimingsStore.cs
```diff
@@ -0,0 +1,141 @@
...
+                }
+            }
+        }
+
+        public EndpointTimings[] GetTimings(DateTime now)
+        {
+            var result = new List<EndpointTimings>();
+            var intervals = GenerateIntervalIds(now, NumberOfHistoricalIntervals);
+
+            foreach (var endpointName in timings.Keys)
```
Scooletz: With the proposal above, scanning for values is easier as the linked list of chunks provides a natural skip-list based on time. Yes, there could be overlapping chunks (from different instances) but still, you need to scan chunk start/end to qualify it for data retrieval.

mikeminutillo:
I spoke to @tmasternak about potentially introducing a circular buffer for the stored results. This works really well if the endpoint can give us a sequence number for the messages. Then we can just store them at `SequenceNumber` MOD `RecordsKeptPerEndpointInstance`

PR #25, opened 2017-06-14
[WIP] Provide historic data to SP

timbussmann:
Simple (hacked) attempt to store historic data for the UI. The consumer parses the incoming data and selects the data which is intended to be shared with the UI. This data is then stored in a timely ordered collection per endpoint for faster read access.

This is a first draft and lacks several optimizations (e.g. truncating data and maintaining different time resolutions for the provided metrics).

ping @WojcikMike @WilliamBZA @tmasternak


src/ServiceControl.Monitoring/Raw/RawDataProvider.cs
```diff
@@ -1,29 +1,83 @@
...
-        public const string Name = "Raw";
-
-        /// <summary>
-        /// Consumes a new portion of data.
-        /// </summary>
         public void Consume(IReadOnlyDictionary<string, string> headers, JObject data)
         {
-            var name = headers.GetOriginatingEndpoint();
-            contexts.AddOrUpdate(name, data, (context, currentData) => data);
+            var endpointName = headers.GetOriginatingEndpoint();
```
WilliamBZA: Are these specific to diagramming or is this indeed still "raw" data?

Scooletz:
@timbussmann Could you provide some description/link for this PR?

src/ServiceControl.Monitoring/Raw/RawDataProvider.cs
```diff
@@ -1,29 +1,83 @@
...
+            var processingTime = processingTimeMetrics["Histogram"]["Mean"].Value<float>();
+
+            // ordered collections are not threadsafe
+            var endpointLock = endpointLocks.GetOrAdd(endpointName, new object());
+
+            lock (endpointLock)
+            {
+                if (!criticalTimeValues.ContainsKey(endpointLock))
+                    criticalTimeValues[endpointLock] = new SortedList<DateTime, float>();
+                criticalTimeValues[endpointLock].Add(timestamp, criticalTime);
```
Scooletz: I can see adding values but I can't see trimming. This will create ever growing collection.

src/ServiceControl.Monitoring/Raw/RawDataProvider.cs
```diff
@@ -1,29 +1,83 @@
...
+                var metrics = new EndpointMetrics(endpoint.Key);
+
+                var endpointLock = endpoint.Value;
+                lock (endpointLock)
+                {
+                    // check because of potential race condition
+                    if (criticalTimeValues.ContainsKey(endpointLock))
+                    {
+                        var endpointCriticalTime = criticalTimeValues[endpointLock];
+                        metrics.CriticalTime = endpointCriticalTime.SkipWhile(x => x.Key < historicMax).Select(x => x.Value).ToArray();
```
Scooletz: If we `SkipWhile` in here and then select values, maybe it'd be easier to maintain a regular `List<T>` of entries and filter them out in here?

src/ServiceControl.Monitoring/Raw/RawDataProvider.cs
```diff
@@ -1,29 +1,83 @@
...
+
+                result.Add(metrics);
+            }
+
+            return result;
         }
 
-        ConcurrentDictionary<string, JObject> contexts = new ConcurrentDictionary<string, JObject>();
+        readonly ConcurrentDictionary<string, object> endpointLocks = new ConcurrentDictionary<string, object>();
+        readonly Dictionary<object, SortedList<DateTime, float>> criticalTimeValues = new Dictionary<object, SortedList<DateTime, float>>();
```
Scooletz: If we could afford storing k-last entries, instead of a `SortedList<DateTime,float>` one could provide for each endpoint a circullar buffer with an index per endpoint. This would enable lockfree, Interlocked based operations. This might be not needed though (requires knowledge about frequency of operations etc etc)

src/ServiceControl.Monitoring/Raw/RawDataProvider.cs
timbussmann: @Scooletz yes, this isn't addressed yet. In general there needs to be further aggregation of received values for further usage. E.g. I was thinking about multiple collections which "feed into each other" (e.g. aggregate values of the first collection every 5 minutes, feed that single value in the level 2 collection and then truncate the first one). This would provide all the data points to visualize different time intervals with little memory usage (but wouldn't allow to "zoom" on past timespans).

src/ServiceControl.Monitoring/Raw/RawDataProvider.cs
timbussmann: in what regards would that be easier? (except from insert performance)

src/ServiceControl.Monitoring/Raw/RawDataProvider.cs
timbussmann: might be an optimization to consider for later on. This is a first spike to provide historic data to the UI.

src/ServiceControl.Monitoring/Raw/RawDataProvider.cs
Scooletz: except from insert performance which is `O log(n)` I suppose, there's no difference

src/ServiceControl.Monitoring/Raw/RawDataProvider.cs
timbussmann: @WilliamBZA what are you referring to?

src/ServiceControl.Monitoring/Raw/RawDataProvider.cs
Scooletz: I like the idea of leveled storage for this. I'd say, that for now, if there's none, the count should be checked and trimmed if needed.

src/ServiceControl.Monitoring/Raw/RawDataProvider.cs
timbussmann: in case new data is appended at the end of a sortedList, insert performance is quite fast afaik. As new data in most cases should be inserted at the end (because it has the latest timestamp), this should work fairly well but also handle edge cases (out of order messages) fairly well as the performance on inserting is not that critical (it's just a message handler) while lookup performance while querying might be more important.

timbussmann: ah, didn't dream it:

> Insertion and removal are generally O(n); however, insertion is O(1) for data that are already in sort order, so that each element is added to the end of the list. (This assumes that a resize is not required.)

https://docs.microsoft.com/en-us/dotnet/standard/collections/sorted-collection-types

didn't verify the source code though.

Scooletz: Good finding! I think I must have thought of SortedDictionary or something else. Or maybe I dreamt it. Anyway, this means that as we are highly likely to insert in at the end therefore, use O(1)

timbussmann:
closing this, replaced by #21

PR #26, opened 2017-07-05
Use content negotiation instead of text

WilliamBZA:
Changes the serialization to use proper json serialization instead of the AsText


WilliamBZA:
Closing as the metrics is being reworked

PR #28, opened 2017-07-12
[WIP] Add additional concurrent dictionary for instances

WilliamBZA:


src/ServiceControl.Monitoring/Timings/EndpointStoreSummarizer.cs
```diff
@@ -0,0 +1,65 @@
...
+                    .ToArray();
+
+                endpointsResult.Add(new MonitoredEndpoint
+                {
+                    Name = endpointName,
+                    EndpointInstanceIds = endpointInstanceIds,
+                    ProcessingTime = new MonitoredEndpointTimings
+                    {
+                        Average = processingTime?.Average,
+                        Points = processingTime?.EndpointInstanceTimings.SelectMany(instance => instance.Intervals).OrderBy(i => i.IntervalStart).Select(i => i.Average).ToArray()
```
tmasternak: Won't this result in multiple points for the same intervals?

src/ServiceControl.Monitoring/Timings/EndpointStoreSummarizer.cs
```diff
@@ -0,0 +1,65 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace ServiceControl.Monitoring.Timings
+{
+    class EndpointStoreSummarizer
+    {
+        public static List<MonitoredEndpoint> BuildMonitoredEndpointSummary(ProcessingTimeStore processingTimeStore, CriticalTimeStore criticalTimeStore)
```
tmasternak: Would it be easier if the `TimingStore` returned per physical instance intervals without aggregation and we had a separate logic that would do grouping by logical name and calculate averages? So the approach would be:
 1. Query `TimingStore`
 2. Group by logical name
 3. In each group group by intervalStart (suming totalAvg and totalMeasurements)
 4. Calculate averages for interval

For instances we would do:
 1. Query `TimingStore`
 2. Filter by physicalIds
 3. Calculate averages for intervals

PR #29, opened 2017-07-12
Adding queue length diagram

WojcikMike:

## Overview
This PR introduces logic for calculating queue length metric value per logical endpoints and remote interface for `ServicePulse` that enables querying calculated values.

### Data gathering
The value of queue length for logical endpoint is a sum of lengths of all input queues. At the endpoint level the following data is gathered and later sent to `SC.Monitoring` instance:
  * Each logical destination is assigned a unique `session-key` and a counter incremented with each outgoing message. `session-key` and `send-counter` are stored in headers of all outgoing messages
  * Per each physical queue all unique received values of `session-key` are stored as well as `receive-counter` which is the highest value of `send-counter` seen for all received message with given `session-key` 

Periodically (`session-key`, `send-counter`) pairs and (`queue-name`, `session-key`, `receive-counter`) tuples are sent to the `SC-Monitoring` instance.

### Data Processing
At the high level queue length calculation works by processing report messages received and updating time based entries for logical endpoints. `SC.Monitoring` instance stores two types of data:
  * Most up-to-date (`session-key`, `send-counter`) and (`queue-name`, `session-key`, `receive-counter`) for all monitored endpoints
  * `Dictionary<string, Dictionary<DateTime, long>>` for each logical endpoint. 

When a report message is received it is processed in following steps:
  * (`session-key`, `send-counter`) are (`queue-name`, `session-key`, `receive-counter`) updated and based on that a list of logical endpoints is build for which a new most up to date queue length estimation can be calculated,
  * The new queue length estimation is calculated for logical endpoints identified in the previous point and stored in `Dictionary<string, Dictionary<DateTime, long>>`


src/ServiceControl.Monitoring.AcceptanceTests/Tests/When_querying_timings_data.cs
```diff
@@ -25,22 +23,15 @@ public class When_querying_timings_data : NServiceBusAcceptanceTest
...
         {
-            using (var client = new HttpClient())
-            {
-                client.DefaultRequestHeaders
-                    .Accept
-                    .Add(new MediaTypeWithQualityHeaderValue("application/json"));
-
-                return client.GetStringAsync("http://localhost:1234/monitored-endpoints").GetAwaiter().GetResult();
-            }
+            return client.GetStringAsync("http://localhost:1234/monitored-endpoints").GetAwaiter().GetResult();
```
Scooletz: Again, possible extraction of `GetString` helper method in the base test class.

src/ServiceControl.Monitoring/Http/Diagrams/TimingsAggregator.cs
```diff
@@ -21,22 +25,35 @@ public IEnumerable<MonitoredEndpoint> AggregateIntoLogicalEndpoints()
...
+            var queueLengths = queueLengthDataStore.GetQueueLengths(now);
 
-            var endpointNames = processingTimes.Concat(criticalTimes).Select(t => t.Id.EndpointName).Distinct();
+            var endpointNames = processingTimes.Concat(criticalTimes).Select(t => t.Id.EndpointName).Concat(queueLengths.Keys).Distinct();
 
             foreach (var endpointName in endpointNames)
             {
                 var instanceProcessingTime = processingTimes.Where(pt => pt.Id.EndpointName == endpointName).ToList();
                 var instanceCriticalTime = criticalTimes.Where(ct => ct.Id.EndpointName == endpointName).ToList();
+                var queueLength = queueLengths.FirstOrDefault(ql => ql.Key == endpointName);
```
Scooletz: _This comment is not only for this PR. Feel free to mark it as a separate issue._

I've just noticed that we're scanning over the collections to search for a specific item. This means complexity of O (n^2). If that's the case, finding by the endpoint name, should find better structures (dicts) to find these?

src/ServiceControl.Monitoring/Http/Diagrams/TimingsAggregator.cs
```diff
@@ -21,22 +25,35 @@ public IEnumerable<MonitoredEndpoint> AggregateIntoLogicalEndpoints()
...
                     EndpointInstanceIds = new[] {instanceProcessingTime, instanceCriticalTime }
                         .SelectMany(t => t).Select(t => t.Id.InstanceId)
                         .Distinct().ToArray(),
                     ProcessingTime = AggregateTimings(instanceProcessingTime),
-                    CriticalTime = AggregateTimings(instanceCriticalTime)
+                    CriticalTime = AggregateTimings(instanceCriticalTime),
+                    QueueLength = new LinearMonitoredValues()
+                    {
+                        Average = queueLength.Value?.Values.Average(),
+                        Points = queueLength.Value?.OrderBy(kvp => kvp.Key)
```
Scooletz: `queueLength.Value?.OrderBy(kvp => kvp.Key)` can be extracted to a variable and `ToArray`ed as it's used in `Points` as in `PointsAxisValues`.

src/ServiceControl.Monitoring/QueueLength/QueueLengthCalculator.cs
```diff
@@ -3,45 +3,81 @@
...
+                double sent;
+                sentSequences.TryGetValue(virtualQueueId.SessionKey, out sent);
 
-                queueLength += sent;
-                queueLength -= received;
+                var received = receivedSequences[virtualQueueId];
 
-                queueLength = Math.Max(0, queueLength);
-                queueLengths[queueName] = queueLength;
+                //TODO: make sure that if we get two reports for the same virtual queue we keep the max values only
```
Scooletz: What's with this TODO?

Scooletz: Should this `TODO` get addressed?

src/ServiceControl.Monitoring/QueueLength/QueueLengthCalculator.cs
```diff
@@ -3,45 +3,81 @@
...
+            return sentSequences.ContainsKey(key);
         }
 
-        public void UpdateSentSequence(string key, long value)
+        public List<VirtualQueueId> UpdateSentSequence(string key, double value)
         {
-            sentSequences.AddOrUpdate(key, _ => value, (_, prev) => Math.Max(prev, value));
+            sentSequences.AddOrUpdate(key, _ => value, (_, previousValue) => Math.Max(previousValue, value));
+
+            return receivedSequences.Keys.Where(k => k.SessionKey == key).ToList();
```
Scooletz: Whenever `UpdateSendSequence` is called, whole `receivedSequences` concurrent dictionary is locked because of access to `Keys` collections. Is it ok?
Should we keep a separate concurrent lookup to address the query by `k.SessionKey == key`?
Should we separate the query from the command and provide another method to query by the `key`?

src/ServiceControl.Monitoring/QueueLength/QueueLengthCalculator.cs
```diff
@@ -3,45 +3,81 @@
...
         }
 
-        public Dictionary<string, long> GetQueueLengths()
+        public Dictionary<VirtualQueueId, double> GetQueueLengths()
         {
-            var queueLengths = new Dictionary<string, long>();
+            var queueLengths = new Dictionary<VirtualQueueId, double>();
 
-            foreach (var kvp in sessionKeyToQueueMapping)
+            foreach (var virtualQueueId in receivedSequences.Keys)
```
Scooletz: After going through the implementation of a concurrent dictionary, I'd say, that we should use `foreach` over the dictionary rather than accessing it's keys to get the query. Creating an enumerator over concurrent dictionary uses just volatile reads, without issuing any locks. As far as I see it, it'd be suitable for this situation. Additionally, we'd receive both the key and the value in one read.

src/ServiceControl.Monitoring/QueueLength/QueueLengthCalculator.cs
```diff
@@ -3,45 +3,81 @@
...
+            if (ReferenceEquals(this, obj)) return true;
+            if (obj.GetType() != this.GetType()) return false;
+            return Equals((VirtualQueueId) obj);
+        }
+
+        public override int GetHashCode()
+        {
+            unchecked
+            {
+                var hashCode = (QueueName != null ? QueueName.GetHashCode() : 0);
```
Scooletz: Just thinking, isn't string.Length sufficient enough for this hash? Have you considered other implementations?

src/ServiceControl.Monitoring/QueueLength/QueueLengthDataStore.cs
```diff
@@ -11,22 +13,57 @@ public QueueLengthDataStore(IQueueLengthCalculator calculator)
             this.calculator = calculator;
         }
 
-        public IEnumerable<KeyValuePair<string, JObject>> Current
+        public Dictionary<string, Dictionary<DateTime, double>> GetQueueLengths(DateTime now)
         {
-            get { return calculator.GetQueueLengths().Select(kvp => new KeyValuePair<string, JObject>(kvp.Key, new JObject(new JProperty("Count", kvp.Value)))); }
+            var results = new Dictionary<string, Dictionary<DateTime, double>>();
+
+            foreach (var endpointName in queueLengths.Keys)
```
Scooletz: Iterating over `.Keys` could be replaced by iteration over the concurrent dictionary which does not take locks.

src/ServiceControl.Monitoring/QueueLength/QueueLengthDataStore.cs
```diff
@@ -11,22 +13,57 @@ public QueueLengthDataStore(IQueueLengthCalculator calculator)
...
+            {
+                ConcurrentDictionary<DateTime, double> endpointData;
+
+                if (queueLengths.TryGetValue(endpointName, out endpointData))
+                {
+                    var endpointDataSnapshot = endpointData.ToArray();
```
Scooletz: Could you describe the approach applied in here? Is it "entries fresh enough to be considered + a cleanup phase"? I think I need a better understanding to yay/nay it.

src/ServiceControl.Monitoring/QueueLength/QueueLengthDataStore.cs
```diff
@@ -38,12 +75,30 @@ void UpdateSends(IEnumerable<JToken> sends)
...
+
+                allUpdatedVirtualQueues.AddRange(updatedVirtualQueues);
             }
+
+            return allUpdatedVirtualQueues.Select(v => v.EndpointName).Distinct().ToList();
+        }
+
+        void SnapshotValueForEndpoint(string endpointName)
+        {
+            var endpointQueueLength = calculator
```
Scooletz: I think I need to get better understanding of this snapshotting approach. Could you describe it?

src/ServiceControl.Monitoring.AcceptanceTests/Tests/When_querying_queue_length_data.cs
```diff
@@ -25,42 +25,26 @@ public class When_querying_queue_length_data : NServiceBusAcceptanceTest
...
-                                    {
-                                        "Count", 10
-                                    }
-                                }
-                            }
-                        }
-                    )
-                }
-            };
+            var result = JObject.Parse(context.Response);
```
tmasternak: I think that this test was meant to play are role of a smoke test rather than testing the schema of the response. That being said I think it should be replaced by a test that calls `monitored-endpoints` as `metrics/queue-length` is not used at all.

src/ServiceControl.Monitoring/QueueLength/QueueLengthCalculator.cs
tmasternak: This is obsolete. Removed.

src/ServiceControl.Monitoring/Http/Diagrams/TimingsAggregator.cs
tmasternak: I think that we need more data to answer this one. My gut feeling is that we should be ok with the current approach. The reason is that number number `ServicePulse` instances being used at the same time will most likely never exceed more than a couple and number of endpoints in range of 100s.

src/ServiceControl.Monitoring/QueueLength/QueueLengthCalculator.cs
tmasternak: You mean length of all 3 properties?

src/ServiceControl.Monitoring/QueueLength/QueueLengthDataStore.cs
tmasternak: The description has been updated

src/ServiceControl.Monitoring/QueueLength/QueueLengthDataStore.cs
tmasternak: The description of this PR has been updated

src/ServiceControl.Monitoring/QueueLength/QueueLengthCalculator.cs
Scooletz: Or hashed lengths.

tmasternak:
@Scooletz please see the review changes

Scooletz:
Awesome 🎉

WojcikMike:
@mikeminutillo would you be so nice and review this PR and merge it?

src/ServiceControl.Monitoring/Http/Diagrams/TimingsAggregator.cs
```diff
@@ -39,6 +44,23 @@ public IEnumerable<MonitoredEndpoint> AggregateIntoLogicalEndpoints()
                     CriticalTime = AggregateTimings(instanceCriticalTime)
                 };
 
+                var queueLength = queueLengths.FirstOrDefault(ql => ql.Key == endpointName);
```
mikeminutillo: Should this be a TryGet on the dictionary rather than scan for key, check value for null?

src/ServiceControl.Monitoring/Http/Diagrams/TimingsAggregator.cs
```diff
@@ -39,6 +44,23 @@ public IEnumerable<MonitoredEndpoint> AggregateIntoLogicalEndpoints()
...
+                var queueLengthValues = queueLength.Value?.OrderBy(kvp => kvp.Key).ToArray();
+
+                if (queueLengthValues != null)
```
mikeminutillo: Under what circumstances will the collection contain the key but a null value?

src/ServiceControl.Monitoring/Http/Diagrams/TimingsAggregator.cs
```diff
@@ -39,6 +44,23 @@ public IEnumerable<MonitoredEndpoint> AggregateIntoLogicalEndpoints()
...
+                {
+                    var queueLengthMinDate = queueLength.Value.Min(kvp => kvp.Key);
```
mikeminutillo: We just sorted these values above into a collection that we have guaranteed to be non-null. We could just take the first instance as it will be the min.

src/ServiceControl.Monitoring/Http/Diagrams/TimingsAggregator.cs
```diff
@@ -39,6 +44,23 @@ public IEnumerable<MonitoredEndpoint> AggregateIntoLogicalEndpoints()
...
+
+                    monitoredEndpoint.QueueLength = new LinearMonitoredValues
+                    {
+                        Average = queueLength.Value?.Values.Average(),
```
mikeminutillo: Haven't we established that queueLength.Value is non-null by the time we get here?

src/ServiceControl.Monitoring/QueueLength/QueueLengthCalculator.cs
```diff
@@ -3,45 +3,80 @@
...
+    using System.Linq;
 
     class QueueLengthCalculator : IQueueLengthCalculator
     {
-        public void UpdateReceivedSequence(string key, long value, string queue)
+        public bool UpdateReceivedSequence(string endpointName, string key, double value, string queue)
         {
-            receivedSequences.AddOrUpdate(key, _ => value, (_, prev) => Math.Max(prev, value));
-            sessionKeyToQueueMapping.GetOrAdd(key, _ => queue);
+            var virtualQueue = new VirtualQueueId
```
mikeminutillo: Would it make sense to just pass in a VirtualQueueId here? The callers of abstraction know about it because other methods on this class return it.

src/ServiceControl.Monitoring/QueueLength/QueueLengthCalculator.cs
```diff
@@ -3,45 +3,80 @@
...
 
-                queueLength += sent;
-                queueLength -= received;
+                var received = receivedSequence.Value;
 
-                queueLength = Math.Max(0, queueLength);
-                queueLengths[queueName] = queueLength;
+                queueLengths.Add(receivedSequence.Key, Math.Max(0, sent - received));
             }
+
```
mikeminutillo: It might be possible to replace this whole method with a single LINQ statement. Something like

```
(from r in receivedSequences
join s in sentSequeneces on r.Key.SessionKey equals s.Key
select new KeyValuePair<>(r.Key, Math.Max(s?.Value ?? 0 -r.Value, 0)).ToDictionary();
```

src/ServiceControl.Monitoring/QueueLength/QueueLengthDataStore.cs
```diff
@@ -11,22 +13,59 @@ public QueueLengthDataStore(IQueueLengthCalculator calculator)
...
-        public void Store(JObject data)
+        public void Store(string endpointName, JObject data)
         {
-            var counters = (JArray) data["Counters"] ?? EmptyArray;
-            UpdateSends(counters);
+            var updatedEndpoints = new List<string>();
 
-            var gauges = (JArray) data["Gauges"] ?? EmptyArray;
-            UpdateReceives(gauges);
+            var counters = (JArray) data["Counters"] ?? new JArray();
```
mikeminutillo: I don't like that this class knows about the message format. This is all going to quietly fail if we update the other end.

src/ServiceControl.Monitoring/QueueLength/QueueLengthDataStore.cs
```diff
@@ -11,22 +13,59 @@ public QueueLengthDataStore(IQueueLengthCalculator calculator)
...
+            updatedEndpoints.AddRange(UpdateSends(counters));
+
+            var gauges = (JArray) data["Gauges"] ?? new JArray();
+            updatedEndpoints.AddRange(UpdateReceives(endpointName, gauges));
+
+            foreach (var updatedEndpoint in updatedEndpoints.Distinct())
+            {
```
mikeminutillo: It feels like there is a lot of effort to ensure that we only snapshot things that have updated but it makes the whole thing much harder to read. Are we better off accepting updates as they come in and snapshotting everything on a timer?

mikeminutillo: We're going through something similar with heartbeats at the moment.

src/ServiceControl.Monitoring/QueueLength/QueueLengthDataStore.cs
```diff
@@ -38,12 +77,30 @@ void UpdateSends(IEnumerable<JToken> sends)
...
+                allUpdatedVirtualQueues.AddRange(updatedVirtualQueues);
             }
+
+            return allUpdatedVirtualQueues.Select(v => v.EndpointName).Distinct().ToList();
+        }
+
+        void SnapshotValueForEndpoint(string endpointName)
+        {
+            var endpointQueueLength = calculator
+                .GetQueueLengths()
```
mikeminutillo: Would it make more sense to call this once and then pass the result around to snapshot each individual endpoint?

src/ServiceControl.Monitoring/QueueLength/QueueLengthReportHandler.cs
```diff
@@ -15,7 +15,9 @@ public QueueLengthReportHandler(QueueLengthDataStore queueLengthDataStore)
 
         public Task Handle(MetricReport message, IMessageHandlerContext context)
         {
-            queueLengthDataStore.Store(message.Data);
+            var endpointInstanceId = EndpointInstanceId.From(context.MessageHeaders);
+
+            queueLengthDataStore.Store(endpointInstanceId.EndpointName, message.Data);
```
mikeminutillo: Would it make sense to pass in the full endpoint instance id and have the Queue Length Data Store determine that it only cares about the name? That brings it in line with what the other stores do.

src/ServiceControl.Monitoring/Http/Diagrams/TimingsAggregator.cs
tmasternak: Good point. I don't think it is possible. The collection can be empty but not null.

src/ServiceControl.Monitoring/QueueLength/QueueLengthCalculator.cs
tmasternak: I like it. Changed to OO-Linqu to keep it consistent with the rest of the code-base.

src/ServiceControl.Monitoring/QueueLength/QueueLengthDataStore.cs
tmasternak: Can we address it in separate PR? This is pretty much the way it was before changes made here.

src/ServiceControl.Monitoring/QueueLength/QueueLengthDataStore.cs
tmasternak: Personally I would keep it as is for now. 

From the initial discussions inside TF we want to push the queue-length calcualations in the direction that would enable us to store it in similar (if not identical) way as timings.

tmasternak:
> It took me a while to figure out what was happening with the update/snapshot cycle and I wonder if we can reduce the complexity without sacrificing too much performance. Especially as we're going to do a perf pass anyway

I would like to propose we tackle that when doing work of different historical time periods. One of the ideas to align data generated for queue length with that generated from timings. I think that would be a good moment to simplify current approach.

tmasternak:
@mikeminutillo the review changes are ready.

Scooletz:
>> It took me a while to figure out what was happening with the update/snapshot cycle and I wonder if we can reduce the complexity without sacrificing too much performance. Especially as we're going to do a perf pass anyway

> I would like to propose we tackle that when doing work of different historical time periods. One of the ideas to align data generated for queue length with that generated from timings. I think that would be a good moment to simplify current approach.

The speed and beauty of a good data structure is incoming and is inevitable 😀

PR #30, opened 2017-07-13
Storing timing data per endpoint instance

tmasternak:

### Goal
Store information about number of physical instances for given logical endpoint.

### PoA
 - [x] Parse `instanceId` passed from `NServiceBus.Metrics`
 - [x] Store information about known physical instances 
 - [x] Expose number of physical instances in data returned form API 
 - [x] Update ATT to newest version of `NServiceBus.Metrics`

/cc: @WilliamBZA


src/ServiceControl.Monitoring/EndpointInstanceId.cs
```diff
@@ -0,0 +1,69 @@
...
+
+            var hostId = headers[Headers.OriginatingHostId];
+            var hostDisplayName = headers[Headers.HostDisplayName];
+
+            return new EndpointInstanceId(endpointName, hostId, hostDisplayName);
+        }
+
+        protected bool Equals(EndpointInstanceId other)
+        {
+            return string.Equals(EndpointName, other.EndpointName) &&
```
WilliamBZA: Is it not enough here to simply check the Instance Id?

If 2 instances have the same endpoint name, the only difference will be the instance Id. If they have a different endpoint name, the instance Id will also be different.

src/ServiceControl.Monitoring/EndpointInstanceId.cs
```diff
@@ -0,0 +1,69 @@
...
+
+            return Equals((EndpointInstanceId) obj);
+        }
+
+        public override int GetHashCode()
+        {
+            unchecked
+            {
+                var hashCode = (EndpointName != null ? EndpointName.GetHashCode() : 0);
+                hashCode = (hashCode * 397) ^ (InstanceId != null ? InstanceId.GetHashCode() : 0);
```
WilliamBZA: Same as the equals comparer, can we not simply use the instance Id here?

src/ServiceControl.Monitoring/Http/Diagrams/DiagramApiModule.cs
```diff
@@ -0,0 +1,24 @@
...
+        /// <summary>
+        /// Initializes the metric API module.
+        /// </summary>
+        public MonitoredEndpointsModule(ProcessingTimeStore processingTimeStore, CriticalTimeStore criticalTimeStore)
+        {
+            Get["/monitored-endpoints"] = x =>
+            {
+                var endpointsData = TimingsAggregator.AggregateIntoLogicalEndpoints(processingTimeStore, criticalTimeStore);
+
+                return Negotiate.WithModel(endpointsData.ToArray());
```
WilliamBZA: Don't need the `ToArray()` here.

src/ServiceControl.Monitoring/Http/Diagrams/TimingsAggregator.cs
```diff
@@ -0,0 +1,72 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace ServiceControl.Monitoring.Timings
+{
+    class TimingsAggregator
+    {
+        public static List<MonitoredEndpoint> AggregateIntoLogicalEndpoints(ProcessingTimeStore processingTimeStore, CriticalTimeStore criticalTimeStore)
```
WilliamBZA: I wonder if this code couldn't look a bit better by changing the return type to `IEnumerable` and then `yield return`ing the MonitoredEndpoint instance instead of adding to a full list

src/ServiceControl.Monitoring/EndpointInstanceId.cs
mikeminutillo: I think it should be possible to host 2 endpoints in the same host. That can lead to a situation where two endpoints have different names but the same instance id.

src/ServiceControl.Monitoring/Http/Diagrams/TimingsAggregator.cs
```diff
@@ -0,0 +1,72 @@
...
+        {
+            var now = DateTime.UtcNow;
+
+            var processingTimes = processingTimeStore.GetTimings(now);
+            var criticalTimes = criticalTimeStore.GetTimings(now);
+
+            var endpointNames = new[] { processingTimes, criticalTimes }
+                .SelectMany(t => t).Select(t => t.Id.EndpointName)
```
mikeminutillo: Can the `SelectMany` and `Select` be collapsed into a single `SelectMany`? Either that or a `processingTimes.Concat(criticalTimes).Select(t => t.Id.EndpointName).Distinct()` could do the job.

src/ServiceControl.Monitoring/QueueLength/QueueLengthDataStore.cs
```diff
@@ -16,7 +16,7 @@ public QueueLengthDataStore(IQueueLengthCalculator calculator)
             get { return calculator.GetQueueLengths().Select(kvp => new KeyValuePair<string, JObject>(kvp.Key, new JObject(new JProperty("Count", kvp.Value)))); }
         }
 
-        public void Store(JObject data)
+        public void Store(string endpointName, string instanceId, JObject data)
```
mikeminutillo: Why is this needed? These values don't get used.

src/ServiceControl.Monitoring/EndpointInstanceId.cs
tmasternak: > I think it should be possible to host 2 endpoints in the same host. That can lead to a situation where two endpoints have different names but the same instance id.

That's a good point. Did not realize that 👍

src/ServiceControl.Monitoring/Http/Diagrams/TimingsAggregator.cs
tmasternak: 👍 readability for the way.

src/ServiceControl.Monitoring/QueueLength/QueueLengthDataStore.cs
tmasternak: Removed

src/ServiceControl.Monitoring/Http/Diagrams/TimingsAggregator.cs
tmasternak: I like the `Concat` it reads a lot better. 👍

tmasternak:
@mikeminutillo I've addressed review comments

tmasternak:
Can we merge this one?

PR #33, opened 2017-07-14
[WIP] Add different scaling options

WilliamBZA:
Adds the xaxis values.

There's a bunch of hacks to get some sample data showing.

The key part is that The `Points` array should be a 1-to-1 match to the `PointsAxisValues` array

So a uniform example (this would simulate the current assumption that 1 point = 1 value on the xaxis) would be:

```
Points: [0, 1, 4, 6, 8]
PointsAxisValues: [0, 1, 2, 3, 4]
```

While a non-uniform example would be:

```
Points: [0, 1, 4, 6, 8]
PointsAxisValues: [0, 1, 6, 7, 14]
```

This example implies that there should be linear interpolation between the second point (1) and the third point (4) such that the x-axis shows a gap of 5 spaces between them.


WojcikMike:
PR rebased to develop

WojcikMike:
Ready to be merged

tmasternak:
Closing in favour of #29

PR #34, opened 2017-07-17
More memory friendly time series processing

Scooletz:
This PR introduces a few changes that are a quick wins in terms of performance of time series reporting we do for critical/processing time. The changes include:
- moving more calculations/logic to the deserialization of `LongValueOccurrences` messages
- making `LongValueOccurrences` poolable with the new class `LongValueOccurrences.Pool` (currently, based on a concurrent stack) to reduce possible memory footprint from digesting a lot of messages
- changing the `TimingStore` from `ConcurrentDictionary of ConcurrentDictionaries` to a simpler `Measurement` class that is held for every endpoint and is capable of storing and performing calculations in-situ (a circular buffer approach is used + epochs)
- introducing coarse-grained locking for `Measurement`. As every message is now limited to N entries, using a coarse-grained `by endpoint` locking looks a bit better than using N `AddOrUpdate` operations on the concurrent dictionary. The update part, as it's pure computation shouldn't take long.


src/ServiceControl.Monitoring/Metrics/Raw/LongValueOccurrences.cs
```diff
@@ -1,23 +1,63 @@
...
+            }
 
-        public long[] Values;
-    }
+            public void Release(LongValueOccurrences message)
+            {
+                message.index = InitialIndex;
+                Array.Clear(message.entries, 0, MaxEntries);
+                pool.Push(message);
+            }
```
mikeminutillo: Should there just be a `Clear()` method on `LongValueOccurrence`?

src/ServiceControl.Monitoring/Metrics/Raw/LongValueOccurrences.cs
```diff
@@ -1,23 +1,63 @@
...
+        }
```
mikeminutillo: Should we limit the size of the pool?

src/ServiceControl.Monitoring/Metrics/Raw/LongValueOccurrences.cs
```diff
@@ -1,23 +1,63 @@
...
+            entries[index].DateTicks = dateTicks;
+            entries[index].Value = value;
+
+            index += 1;
+
+            return true;
+        }
+
+        public long MinDateTick => entries[0].DateTicks;
+        public long MaxDateTick => entries[index].DateTicks;
```
mikeminutillo: This assumes the values are in order. Is that a safe assumption?

src/ServiceControl.Monitoring/Metrics/Raw/LongValueOccurenceSerializerDefinition.cs
```diff
@@ -24,33 +24,52 @@ public void Serialize(object message, Stream stream)
...
+                {
+                    var ticks = reader.ReadInt32();
+                    var value = reader.ReadInt64();
+                    var date = baseTicks + ticks;
 
-            message.Ticks = new int[count];
-            message.Values = new long[count];
+                    if (message.TryRecord(date, value) == false)
+                    {
+                        message = LeaseMessage(messages);
```
mikeminutillo: I don't like that this is asymmetric. We lease new instances in the deserialization step and then release them in a handler. 

Would it be better to introduce a behavior to lease a new message, fill it, call the inner behavior, and then release it, all withing a single try-finally?

mikeminutillo: I guess that the abstractions aren't really set up for this.

src/ServiceControl.Monitoring/Timings/TimingsStore.cs
```diff
@@ -3,122 +3,162 @@
...
+                try
+                {
+                    for (var i = 0; i < message.Length; i++)
                     {
-                        Id = instanceId,
-                        TotalMeasurements = totalMeasurements,
-                        TotalTime = totalDuration,
-                        Intervals = timingIntervals
-                    });
+                        Report(ref message.entries[i]);
```
mikeminutillo: Why are these passed in as `ref`? We don't change them.

src/ServiceControl.Monitoring/Timings/TimingsStore.cs
```diff
@@ -3,122 +3,162 @@
...
+                {
+                    return $"{nameof(TotalMeasurements)}: {TotalMeasurements}, {nameof(Epoch)}: {Epoch}, {nameof(TotalTime)}: {TotalTime}";
+                }
+            }
 
-        struct MeasurementInterval
-        {
-            public int TotalMeasurements;
-            public long TotalTime;
+            static long GetEpoch(ref LongValueOccurrences.Entry entry)
```
mikeminutillo: Would it make sense to move this to LongValueOccurrences.Entry and pass in the interval size?

src/ServiceControl.Monitoring/Timings/TimingsStore.cs
```diff
@@ -3,122 +3,162 @@
...
+                            intervalsToFill[i].TotalTime = interval.TotalTime;
+                            intervalsToFill[i].TotalMeasurements = interval.TotalMeasurements;
 
-                            timingIntervals[i].TotalTime = bucket.TotalTime;
-                            timingIntervals[i].TotalMeasurements = bucket.TotalMeasurements;
+                            totalDuration += interval.TotalTime;
+                            totalMeasurements += interval.TotalMeasurements;
                         }
+
+                        epoch -= 1;
```
mikeminutillo: Should this be `% Size` as well?

src/ServiceControl.Monitoring/Timings/TimingsStore.cs
```diff
@@ -3,122 +3,162 @@
...
-            }
+                long totalDuration;
+                long totalMeasurements;
+                var timingIntervals = new TimingInterval[NumberOfHistoricalIntervals];
 
-            if (instanceData.Count > 2 * NumberOfHistoricalIntervals)
-            {
-                var historySize = TimeSpan.FromTicks(IntervalSize.Ticks * NumberOfHistoricalIntervals);
-                var oldestValidInterval = now.Subtract(historySize).RoundDownToNearest(IntervalSize);
+                measurement.ReportTimeIntervals(now, timingIntervals, out totalDuration, out totalMeasurements);
```
mikeminutillo: Would it make sense to create an instance of `EndpointInstanceTimings` and pass it in to be filled in (with totals)?

src/ServiceControl.Monitoring/Metrics/Raw/LongValueOccurrences.cs
Scooletz: I can extract it :)

src/ServiceControl.Monitoring/Metrics/Raw/LongValueOccurrences.cs
Scooletz: The pool size is effectively limited by the concurrency limit of the transport multiplied by the number of `LongValueOccurences` messages that a single message can be split into. I cannot see any reason to put another limitation on top of it. Is there any?

src/ServiceControl.Monitoring/Metrics/Raw/LongValueOccurrences.cs
Scooletz: That's the format currently used by `NSB.Metrics`. Values are sorted in a single message before sending them out to `SC.Monitoring`. I don't know if there's a test for it, but I can introduce one if there's none to make it a part of the contract.

src/ServiceControl.Monitoring/Metrics/Raw/LongValueOccurenceSerializerDefinition.cs
Scooletz: I see. Let me take a look at this and got back in here. I like your proposal with a behavior.

src/ServiceControl.Monitoring/Timings/TimingsStore.cs
Scooletz: Interesting! Will apply!

src/ServiceControl.Monitoring/Timings/TimingsStore.cs
Scooletz: I don't think so. The `epoch` is always quite big (ticks from zero-day divided by interval). The `%` is applied a few lines above to calculate the epoch index.

```
var epochIndex = epoch % Size;
var interval = intervals[epochIndex];
```

src/ServiceControl.Monitoring/Timings/TimingsStore.cs
Scooletz: Entries are `structs` to make them allocated as a single table. I'm using `ref` to pass a "reference to a struct in a table".

src/ServiceControl.Monitoring/Timings/TimingsStore.cs
Scooletz: The `epochs` are needed only for the `TimingStore`. I see no value in moving this to the `Entry` as `Entry` has no notion of `epoch`.

src/ServiceControl.Monitoring/Timings/TimingsStore.cs
mikeminutillo: Ah correct. The epoch always counts down, the index gets modded. I had them muddled in my head.

src/ServiceControl.Monitoring/Metrics/Raw/LongValueOccurenceSerializerDefinition.cs
danielmarbach: @mikeminutillo I agree it is a bit unfortunate that there is no symmetry but I'm afraid there is not much we can do here since the serializer doesn't allow to float things from the pipeline into the serializer. In order to have symmetry, you'd need a behavior which gets the lease and then passes that into the serializer and then releases the lease after the pipeline returned. I see no way without ambient context to achieve it. So instead of introducing an ambient context I would stay with the current approach.

src/ServiceControl.Monitoring/Metrics/Raw/LongValueOccurenceSerializerDefinition.cs
```diff
@@ -24,33 +24,52 @@ public void Serialize(object message, Stream stream)
...
+                var baseTicks = reader.ReadInt64();
+                var count = reader.ReadInt32();
 
-        static void ToLongValueOccurrences(LongValueOccurrences message, Stream stream)
-        {
-            var reader = new BinaryReader(stream);
+                if (count == 0)
+                {
+                    // no records, return empty
+                    return new object[0];
```
danielmarbach: static empty array? In the future we can finally use Array.Empty

src/ServiceControl.Monitoring/Metrics/Raw/LongValueOccurenceSerializerDefinition.cs
danielmarbach: Another thought. Maybe we are operating on the wrong abstraction. I looks like we envy to be a raw endpoint like a satellite that operates on the main queue and skip all the rest of core magic.

src/ServiceControl.Monitoring.Tests/Timings/TimingsStoreTests.cs
```diff
@@ -36,6 +36,20 @@ public void Number_of_intervals_is_constant_per_know_endpoint()
...
             Assert.AreEqual(TimingsStore.NumberOfHistoricalIntervals, timings[0].Intervals.Length);
+
+            // ordering of intervals
+            var dateTimes = timings[0].Intervals.Select(i => i.IntervalStart).ToArray();
+            var orderedDateTimes = dateTimes.OrderByDescending(d => d).ToArray();
+
+            CollectionAssert.AreEqual(orderedDateTimes, dateTimes);
+
+            // length of intervals
+            var intervalLength = dateTimes[0] - dateTimes[1];
```
tmasternak: How about ? (PS: I could not resist my functional part of my brain ;))
```
Enumerable.Zip(dateTimes, dateTimes.Skip(1), (f, s) => s-f).All(i => i == intervalLength) ?
```

src/ServiceControl.Monitoring/Metrics/Raw/LongValueOccurenceSerializerDefinition.cs
```diff
@@ -24,33 +24,52 @@ public void Serialize(object message, Stream stream)
...
+                    }
+                }
+
+                return messages.ToArray();
             }
+
+            throw new Exception($"The message version number '{version}' cannot be handled properly.");
+        }
+
+        static LongValueOccurrences LeaseMessage(List<object> messages)
```
tmasternak: Are we doing too much here - leasing a message and adding to the collection?

src/ServiceControl.Monitoring/Metrics/Raw/LongValueOccurrences.cs
```diff
@@ -1,23 +1,68 @@
...
+            public LongValueOccurrences Lease()
+            {
+                LongValueOccurrences value;
+                if (pool.TryPop(out value))
+                {
+                    return value;
+                }
 
-        public int[] Ticks;
+                return new LongValueOccurrences();
```
tmasternak: This means that there is no upper bound on the pool size right? I guess we should be fine as this will be throttled by concurrency limit on the endpoint itself.

src/ServiceControl.Monitoring/Timings/TimingsStore.cs
```diff
@@ -3,122 +3,161 @@
...
+            MeasurementInterval[] intervals = new MeasurementInterval[Size];
 
-            foreach (var instanceId in timings.Keys)
+            // ReSharper disable once SuggestBaseTypeForParameter
+            public void ReportTimeIntervals(DateTime now, EndpointInstanceTimings item)
             {
-                ConcurrentDictionary<DateTime, MeasurementInterval> endpointInstanceData;
+                var epoch = GetEpoch(now.Ticks);
+                var intervalsToFill = item.Intervals;
+                var howMany = intervalsToFill.Length;
```
tmasternak: can we call this `numberOfIntervalsToFill` or remove the variable?

src/ServiceControl.Monitoring/Metrics/Raw/LongValueOccurrences.cs
Scooletz: That was my reasoning as well. It's bound by design.

src/ServiceControl.Monitoring/Metrics/Raw/LongValueOccurenceSerializerDefinition.cs
SzymonPobiega: @danielmarbach I think a behavior in the transport receive context would do the trick. I talked about this earlier with @tmasternak but we agreed that serializer was good enough at that stage. As I see, you are trying to squeeze tiny bits of performance in the PRs now. I suggest you drop the serializer and the whole pipeline and use the transport receive context. Or the `NServiceBus.Raw` :-). It is now used in Wormhole gateway, Bridge and the SC.TransportAdapter so you won't be my first customers ;-)

src/ServiceControl.Monitoring.Tests/Timings/TimingsStoreTests.cs
Scooletz: It took me 2 min to read it. I'd stay with the current dummy assertion.

src/ServiceControl.Monitoring/Metrics/Raw/LongValueOccurenceSerializerDefinition.cs
Scooletz: Will inline.

src/ServiceControl.Monitoring/Timings/TimingsStore.cs
Scooletz: Will rename.

Scooletz:
Ping @tmasternak

src/ServiceControl.Monitoring/Metrics/Raw/LongValueOccurenceSerializerDefinition.cs
Scooletz: I'd like to postpone this change to a separate PR. @mikeminutillo @SzymonPobiega ? can we proceed with the current one?

Scooletz:
Ping @mikeminutillo Did I address all the needed changes?

src/ServiceControl.Monitoring.Tests/Timings/TimingsStoreTests.cs
tmasternak: > It took me 2 min to read it. I'd stay with the current dummy assertion.

Fair enough. My functional side of brain is usually wrong :P.

PR #37, opened 2017-07-24
Retries metric

tmasternak:
## Description
This PR introduces retries metrics. It does so by processing `Occurrences` message published by `NSB.Metrics`. In introduces following changes:
 * Deserializer for `Occurrences` messages
 * Handler for `Occurrences` messages that records data in `RetriesStore` based on `IntervalStore` 
 * Renaming `TimingStore` to `IntervalStore`


src/ServiceControl.Monitoring/Metrics/Raw/OccurrenceSerializerDefinition.cs
```diff
@@ -0,0 +1,42 @@
...
+        {
+            return mapper => new OccurrenceSerializer();
+        }
+    }
+
+    public class OccurrenceSerializer : RawMessageSerializer<Occurrences>, IMessageSerializer
+    {
+        public object[] Deserialize(Stream stream, IList<Type> messageTypes = null)
+        {
+            long baseTicks = 0;
```
mikeminutillo: I'm not sure about this being captured in an enclosure. Wondering if we should create a `state` struct and pass that around. It can include the creation of the message pool and set up the base ticks on the starter and then get passed in to each subsequent call.

src/ServiceControl.Monitoring/Retries/Retries.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace ServiceControl.Monitoring
+{
+    public class Retries
```
mikeminutillo: What is this?

src/ServiceControl.Monitoring/Http/Diagrams/DiagramApiModule.cs
```diff
@@ -12,9 +12,9 @@ public class MonitoredEndpointsModule : ApiModule
         /// <summary>
         /// Initializes the metric API module.
         /// </summary>
-        public MonitoredEndpointsModule(ProcessingTimeStore processingTimeStore, CriticalTimeStore criticalTimeStore, QueueLengthDataStore queueLengthDataStore)
+        public MonitoredEndpointsModule(ProcessingTimeStore processingTimeStore, CriticalTimeStore criticalTimeStore, QueueLengthDataStore queueLengthDataStore, RetriesStore retriesStore)
```
mikeminutillo: You should just be able to replace this with a dependency on `TimingsAggregator` and Autofac will put everything together.

src/ServiceControl.Monitoring/Metrics/Raw/OccurrenceSerializerDefinition.cs
mikeminutillo: It turned out to be simpler to just drop the timestamp management into the base class. Both current implementation manage timestamps the same way. 

https://github.com/Particular/ServiceControl.Monitoring/pull/38

src/ServiceControl.Monitoring/Retries/Retries.cs
tmasternak: I have no idea :P

src/ServiceControl.Monitoring/Retries/OccurrenceHandler.cs
```diff
@@ -0,0 +1,41 @@
...
+
+                var messageType = context.MessageHeaders[MetricHeaders.MetricType];
+
+                switch (messageType)
+                {
+                    case RetriesMessageType:
+                        store.Store(endpointInstanceId, message.Entries);
+                        break;
+                    default:
+                        throw new UnknownOccurrenceMessageType(messageType);
```
WilliamBZA: Should we have a custom retry policy so these messages get forwarded straight to the error queue instead of going through FLR+SLR?

WilliamBZA: Or perhaps even discarded?

src/ServiceControl.Monitoring/Metrics/Raw/RawMessage.cs
```diff
@@ -0,0 +1,50 @@
...
+    {
+        public int Length => Index;
+
+        void Clear()
+        {
+            Index = InitialIndex;
+            Array.Clear(Entries, 0, MaxEntries);
+        }
+
+        public Entry[] Entries = new Entry[MaxEntries];
```
WilliamBZA: Much better name :+1:

src/ServiceControl.Monitoring/Metrics/Raw/OccurrenceSerializerDefinition.cs
```diff
@@ -0,0 +1,27 @@
...
+        {
+            return mapper => new OccurrenceSerializer();
+        }
+    }
+
+    public class OccurrenceSerializer : RawMessageSerializer<Occurrences>, IMessageSerializer
+    {
+        public object[] Deserialize(Stream stream, IList<Type> messageTypes = null)
+        {
+            return DeserializeRawMessage(stream, (t, r, m) => m.TryRecord(t));
```
WilliamBZA: My OCD here, but can we give these better names than `t`, `r`, and `m`? Especially since the func's types aren't named.

src/ServiceControl.Monitoring/Metrics/Raw/LongValueOccurenceSerializerDefinition.cs
```diff
@@ -15,58 +15,17 @@ class LongValueOccurrenceSerializerDefinition : SerializationDefinition
...
-            var version = reader.ReadInt64();
-
-            if (version == 1)
-            {
-                var baseTicks = reader.ReadInt64();
-                var count = reader.ReadInt32();
-
-                if (count == 0)
+            return DeserializeRawMessage(stream,
+                (t, r, m) =>
```
WilliamBZA: My OCD here, but can we give these better names than `t`, `r`, and `m`? Especially since the func's types aren't named.

Scooletz: Is there a lambda allocation in here? If yes, could we extract the creation of the delegate and pass an already constructed one?

src/ServiceControl.Monitoring/Metrics/Raw/OccurrenceSerializerDefinition.cs
Scooletz: Is there a lambda allocation in here? If yes, could we extract the creation of the delegate and pass an already constructed one?

src/ServiceControl.Monitoring/Metrics/Raw/RawMessageSerializer.cs
```diff
@@ -0,0 +1,59 @@
...
+
+                var messages = new List<object>(1); // usual case
+
+                T message = null;
+
+                for (var i = 0; i < count; i++)
+                {
+                    if (i % RawMessage.MaxEntries == 0)
+                    {
+                        message = RawMessage.Pool<T>.Default.Lease();
```
Scooletz: Tricky :)

src/ServiceControl.Monitoring/Metrics/Raw/RawMessageSerializer.cs
```diff
@@ -0,0 +1,59 @@
...
+    public class RawMessageSerializer<T> where T : RawMessage, new()
+    {
+        static readonly object[] NoMessages = new object[0];
+
+        public void Serialize(object message, Stream stream)
+        {
+            throw new NotImplementedException();
+        }
+
+        protected object[] DeserializeRawMessage(Stream stream, Func<long, BinaryReader, T, bool> store)
```
Scooletz: `store` could be an `abstract` method of this type (`RawMessageSerializer` is used only as a base class) removing the need of passing the storing delegate.

src/ServiceControl.Monitoring/Timings/IntervalsStore.cs
```diff
@@ -3,33 +3,35 @@
     using System;
     using System.Collections.Concurrent;
     using System.Collections.Generic;
+    using System.Linq;
     using System.Threading;
+    using Infrastructure;
     using Metrics.Raw;
 
-    public abstract class TimingsStore
+    public abstract class IntervalsStore : IKnowAboutEndpoints
```
Scooletz: I'm not a native English speaker but this name is somewhat strange `IntervalsStore`. Shouldn't it be `IntervalStore`?

src/ServiceControl.Monitoring/Retries/RetriesStore.cs
```diff
@@ -0,0 +1,46 @@
+﻿namespace ServiceControl.Monitoring
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using Infrastructure;
+    using Timings;
+    public class RetriesStore : IntervalsStore, IProvideEndpointMonitoringData, IProvideEndpointInstanceMonitoringData
```
Scooletz: I'm not a native English speaker but this name is somewhat strange `RetriesStore`. Shouldn't it be `RetryStore `?

src/ServiceControl.Monitoring/Http/Diagrams/TimingsAggregator.cs
```diff
@@ -111,6 +64,7 @@ public class MonitoredEndpoint
         public string[] EndpointInstanceIds { get; set; }
         public MonitoredEndpointValues ProcessingTime { get; set; }
         public MonitoredEndpointValues CriticalTime { get; set; }
+        public MonitoredEndpointValues Retries { get; set; }
```
Scooletz: So we abstract `IProvideEndpointInstanceMonitoringData` and `IProvideEndpointMonitoringData` implementations but still keep the properties in here, "knowing" what are `Retries`, `ProcessingTime`, `CriticalTime`. It seems incoherent to me. A dictionary-like approach, with implementations of `IProvide...` adding keys, would be better maybe?

src/ServiceControl.Monitoring/Timings/CriticalTimeStore.cs
```diff
@@ -1,7 +1,29 @@
...
+    public class CriticalTimeStore : IntervalsStore, IProvideEndpointMonitoringData, IProvideEndpointInstanceMonitoringData
     {
-        
+        public void FillIn(MonitoredEndpoint[] data, DateTime now)
+        {
+            var snapshot = GetIntervals(now).ToLookup(x => x.Id.EndpointName);
+
+            foreach (var item in data)
+            {
+                item.CriticalTime = AggregateTimings(snapshot[item.Name].ToList());
```
Scooletz: Will this lookup `snapshot` always include `item.Name`?

src/ServiceControl.Monitoring/Timings/ProcessingTimeStore.cs
```diff
@@ -1,7 +1,28 @@
...
+    public class ProcessingTimeStore : IntervalsStore, IProvideEndpointMonitoringData, IProvideEndpointInstanceMonitoringData
     {
-        
+        public void FillIn(MonitoredEndpoint[] data, DateTime now)
+        {
+            var snapshot = GetIntervals(now).ToLookup(x => x.Id.EndpointName);
+
+            foreach (var item in data)
+            {
+                item.ProcessingTime = AggregateTimings(snapshot[item.Name].ToList());
```
Scooletz: Will this lookup `snapshot` always include `item.Name`?

src/ServiceControl.Monitoring/Http/Diagrams/TimingsAggregator.cs
tmasternak: I propose to make this a separate PR in which I will try to make `MonitoredEndpointValue` more generic.

src/ServiceControl.Monitoring/Metrics/Raw/RawMessageSerializer.cs
tmasternak: Done.

src/ServiceControl.Monitoring/Metrics/Raw/LongValueOccurenceSerializerDefinition.cs
tmasternak: This is solved by abstract `Store` method in base class.

tmasternak: Solved by `Store` method in base class.

src/ServiceControl.Monitoring/Metrics/Raw/OccurrenceSerializerDefinition.cs
tmasternak: Solved by Store method in base class.

src/ServiceControl.Monitoring/Timings/CriticalTimeStore.cs
tmasternak: No but in such case an empty list will get created.

src/ServiceControl.Monitoring/Timings/ProcessingTimeStore.cs
tmasternak: No but in such case an empty list will get created.

src/ServiceControl.Monitoring/Timings/IntervalsStore.cs
tmasternak: I would kept the plural to follow `TimingsStore` convention.

src/ServiceControl.Monitoring/Retries/RetriesStore.cs
tmasternak: Same as [above](https://github.com/Particular/ServiceControl.Monitoring/pull/37#discussion_r129402725).

tmasternak:
@Scooletz I've addressed your comments. I think that performance tests and making `MonitoredEndpoint` class more generic should be addressed in separate PRs.

Scooletz:
LGTM @tmasternak

src/ServiceControl.Monitoring/Http/Diagrams/TimingsAggregator.cs
mikeminutillo: Yeah. This is certainly the intent. That change needs to be done in conjunction with the UI though as it will affect the UI.

src/ServiceControl.Monitoring/Timings/ProcessingTimeStore.cs
```diff
@@ -1,7 +1,28 @@
...
+            }
+        }
+
+        public void FillIn(MonitoredEndpointInstance[] data, DateTime now)
+        {
+            var snapshot = GetIntervals(now).ToLookup(k => $"{k.Id.EndpointName}-{k.Id.InstanceId}");
```
mikeminutillo: If the view model included a proper EndpointInstanceId then we could just lookup based on that.

tmasternak:
@mikeminutillo can we merge this?

PR #39, opened 2017-07-25
Apply OCP

mikeminutillo:
Refactoring to allow all new metric data to be added with all new code.

Should never have to edit the Timing Aggregator again.

If we refactor the result objects to contain (endpoint name, dictionary) then we don't even need to edit the results either. Each new metric can simply add it's key in. 

Thoughts?


src/ServiceControl.Monitoring/Infrastructure/IKnowAboutEndpoints.cs
```diff
@@ -0,0 +1,45 @@
...
+    using System.Collections.Generic;
+    using System.Linq;
+    using Timings;
+
+    public interface IKnowAboutEndpoints
+    {
+        IDictionary<string, IEnumerable<EndpointInstanceId>> AllEndpointData();
+    }
+
+    public interface IProvideEndpointMonitoringData
```
tmasternak: Should we extend this and `IProvideEndpointInstanceMonitoringData` with metric name. In that way we can make `MonitoredEndpoint` and `MonitoredEndpointInstance` unaware of the metrics.

src/ServiceControl.Monitoring/Timings/IntervalsStore.cs
```diff
@@ -176,5 +178,28 @@ public class TimeInterval
...
         static TimeSpan IntervalSize = TimeSpan.FromSeconds(15);
+
+        IDictionary<string, IEnumerable<EndpointInstanceId>> IKnowAboutEndpoints.AllEndpointData()
+        {
+            return intervals.Keys
+                .GroupBy(i => i.EndpointName)
+                .ToDictionary(g => g.Key, g => g.AsEnumerable());
+        }
+
+        protected static MonitoredEndpointValues AggregateTimings(List<EndpointInstanceIntervals> timings)
```
tmasternak: Should we move this and `AggregateRetries` to separate static classes calling them `AggregateToValueAverage` and `AggregateToCount`. In the near future we will have message rate which will use the same aggregation as `Retries`

tmasternak:
@mikeminutillo I'm looking into the ATT to figure out why it's failing

src/ServiceControl.Monitoring/Timings/IntervalsStore.cs
mikeminutillo: Either that or rename `AggregateRetries` to `AggregateToCount` and push it down into the timing store. I think we'll want to refactor some of this afterwards anyway.

src/ServiceControl.Monitoring/Infrastructure/IKnowAboutEndpoints.cs
mikeminutillo: Yeah. I think that's the next logical step. Then adding a new one is literally adding an interface to a class.

tmasternak:
@mikeminutillo I propose we merge #37 and this one after it.

PR #40, opened 2017-07-26
Queue length based on interval store

tmasternak:
## Description
This PR changes the way queue-length metric is calculated and stored. Firstly it introduces `QueueLengthSnapshotting` task that is responsible for storing currently know queue length estimations for all endpoints. Snapshot values are stored in `IntervalStore` just like any other metrics.


src/ServiceControl.Monitoring/Http/Diagrams/TimingsAggregator.cs
```diff
@@ -65,7 +65,7 @@ public class MonitoredEndpoint
         public MonitoredEndpointValues ProcessingTime { get; set; }
         public MonitoredEndpointValues CriticalTime { get; set; }
         public MonitoredEndpointValues Retries { get; set; }
-        public LinearMonitoredValues QueueLength { get; set; }
+        public MonitoredEndpointValues QueueLength { get; set; }
```
Scooletz: 👍 Series like everything else 😍

src/ServiceControl.Monitoring/QueueLength/IQueueLengthCalculator.cs
```diff
@@ -4,8 +4,8 @@
 
     public interface IQueueLengthCalculator
     {
-        bool UpdateReceivedSequence(VirtualQueueId virtualQueueId, double value);
-        List<VirtualQueueId> UpdateSentSequence(string key, double value);
+        void UpdateReceivedSequence(VirtualQueueId virtualQueueId, double value);
```
Scooletz: Nice `voids`

src/ServiceControl.Monitoring/QueueLength/QueueLengthDataStore.cs
```diff
@@ -1,117 +1,103 @@
...
+                    {
+                        DateTicks = now.Ticks,
+                        Value = (long)g.Sum(i => i.Value)
+                    }
+                });
+
+            foreach (var queueLength in queueLengths)
             {
-                SnapshotValueForEndpoint(virtualQueueLengths, updatedEndpoint);
+                Store(queueLength.Id, new[]
```
Scooletz: The `IntervalsStore` works in following way. If there is an entry responsible for storing data within the given period it will add the number [to the specific bucket](https://github.com/Particular/ServiceControl.Monitoring/blob/c373b51c8cf760098b413f4ff4c19f00f6d06ff3/src/ServiceControl.Monitoring/Timings/IntervalsStore.cs#L119).

```
intervals[epochIndex].TotalTime += entry.Value;
intervals[epochIndex].TotalMeasurements += 1;
```

Consider a following scenario: the queue length does not change and is reported second after second which fits into one bucket. Then the `Value` will be bumped up multiple times with the same value.

Am I seeing it wrong?

We could check this by snapshotting the value twice in one test and see if it's different.

mikeminutillo: Does this mean that we should not snapshot the values faster than the epoch interval?

Scooletz: I'd say @mikeminutillo that we need to provide a better store, able to either: provide aggregation (like we do for durations) or snapshoting, replacing the whole value. This issue, needs to be verified by a test though.

src/ServiceControl.Monitoring/QueueLength/IQueueLengthCalculator.cs
WilliamBZA: They're the best voids. Everybody says so.

src/ServiceControl.Monitoring/QueueLength/QueueLengthDataStore.cs
tmasternak: The assumption was that proper average over interval will be calculated during [aggregation](https://github.com/Particular/ServiceControl.Monitoring/pull/40/files#diff-6d7c64ccf40a3838f6b3e49073c5d004R69) but here is actually an error there :/. The value for interval should be TotalValue/TotalMeasurements. Will fix.

src/ServiceControl.Monitoring/Infrastructure/EndpointRegistry.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace ServiceControl.Monitoring.Infrastructure
+{
+    using System.Collections.Concurrent;
+    using System.Collections.Generic;
+    using System.Linq;
+
+    public class EndpointRegistry
+    {
+        ConcurrentBag<EndpointInstanceId> endpointInstances = new ConcurrentBag<EndpointInstanceId>();
```
tmasternak: Correct. Switched to dictionary

src/ServiceControl.Monitoring/QueueLength/QueueLengthReportHandler.cs
```diff
@@ -1,22 +1,27 @@
...
         {
+            this.endpointRegistry = endpointRegistry;
             this.queueLengthDataStore = queueLengthDataStore;
         }
 
         public Task Handle(MetricReport message, IMessageHandlerContext context)
         {
             var endpointInstanceId = EndpointInstanceId.From(context.MessageHeaders);
 
+            endpointRegistry.Record(endpointInstanceId);
```
tmasternak: I've added dedicated handler with explicit declaration of messages. I think it will make it less fragile if for some reason we start getting messages which are not metric reports.

src/ServiceControl.Monitoring/QueueLength/QueueLengthDataStore.cs
tmasternak: [Fixed](https://github.com/Particular/ServiceControl.Monitoring/pull/40/files#diff-6d7c64ccf40a3838f6b3e49073c5d004R72)

tmasternak:
@Scooletz @mikeminutillo  please see comments and fixes

src/ServiceControl.Monitoring/QueueLength/QueueLengthDataStore.cs
Scooletz: Thank you for clarifying @tmasternak This was not obvious to me.

src/ServiceControl.Monitoring/QueueLength/QueueLengthDataStore.cs
```diff
@@ -63,16 +63,18 @@ static MonitoredEndpointValues AggregateQueueLength(List<EndpointInstanceInterva
...
             {
-                Average = intervals.Sum(t => t.TotalValue) / returnOneIfZero(intervals.Sum(t => t.Intervals.Length)),
                 Points = intervals.SelectMany(t => t.Intervals)
                     .GroupBy(i => i.IntervalStart)
                     .OrderBy(g => g.Key)
                     .Select(g => g.Sum(i => i.TotalValue / returnOneIfZero(i.TotalMeasurements)))
                     .ToArray()
             };
 
+            result.Average = result.Points.Length > 0 ? result.Points.Average() : 0d;
```
Scooletz: I see a fix applied on a commit that was green on the build server without a test being changed? Are we lacking the coverage in here?

src/ServiceControl.Monitoring/Infrastructure/EndpointTracker.cs
```diff
@@ -0,0 +1,41 @@
+﻿namespace ServiceControl.Monitoring.Infrastructure
+{
+    using System.Threading.Tasks;
+    using Metrics.Raw;
+    using NServiceBus;
+    using NServiceBus.Metrics;
+
+    public class EndpointTracker : IHandleMessages<MetricReport>, IHandleMessages<LongValueOccurrences>, IHandleMessages<Occurrences>
```
Scooletz: Beautiful 🥇

src/ServiceControl.Monitoring/QueueLength/QueueLengthDataStore.cs
```diff
@@ -69,7 +69,7 @@ static MonitoredEndpointValues AggregateQueueLength(List<EndpointInstanceInterva
                 Points = intervals.SelectMany(t => t.Intervals)
                     .GroupBy(i => i.IntervalStart)
                     .OrderBy(g => g.Key)
-                    .Select(g => (double)g.Sum(i => i.TotalValue))
+                    .Select(g => g.Sum(i => i.TotalValue / returnOneIfZero(i.TotalMeasurements)))
```
Scooletz: Was it shown by a failing test?

src/ServiceControl.Monitoring/QueueLength/QueueLengthDataStore.cs
tmasternak: I think we do. I would keep it out of this PR as a proper [test](https://github.com/Particular/ServiceControl.Monitoring/pull/41/files#diff-13946e3bafd2e391e0a09012f7372b51R1) for aggregations are included in #41.

src/ServiceControl.Monitoring/QueueLength/QueueLengthDataStore.cs
tmasternak: No :/. Please see this [comment](https://github.com/Particular/ServiceControl.Monitoring/pull/40#discussion_r129809797).

src/ServiceControl.Monitoring/QueueLength/QueueLengthReportHandler.cs
mikeminutillo: I would prefer to accidentally start tracking something we didn't intend to than to miss tracking something we thought we'd do magically.

src/ServiceControl.Monitoring/Infrastructure/EndpointTracker.cs
mikeminutillo: As I mentioned above, I would rather trap `IMessage` and accidentally track something we didn't mean to, than forget to add a handler in here and miss tracking something in a weird edge case.

tmasternak: @mikeminutillo point taken. Can we make it a separate PR though? I spent some time yesterday trying to make it work but for some reason when switching to `IHandleMessages<IMessage>` or `IHandleMessages<object>` results in `AutoFac` exception - it complains that it cannot find dependencies for the constructor :/.

src/ServiceControl.Monitoring/QueueLength/QueueLengthReportHandler.cs
Scooletz: If we introduced a handler for `IMessage` or `object` we'd be consuming any message. In this case, any misconfigured endpoint sending messages to this queue, would send them into a black hole. Wouldn't it be better to state it in an explicit way, making the subset of accepted messages?

Scooletz:
@mikeminutillo I've merged #41 to this PR (as it was 👍 ), but as this is aimed for `develop` it'd be great if you could review it.

src/ServiceControl.Monitoring/Infrastructure/EndpointRegistry.cs
```diff
@@ -0,0 +1,30 @@
...
+        {
+            endpointInstances.TryAdd(endpointInstanceId, true);    
+        }
+
+        public IDictionary<string, IEnumerable<EndpointInstanceId>> GetAllEndpoints()
+        {
+            return endpointInstances.ToArray()
+                .GroupBy(kv => kv.Key.EndpointName)
+                .ToDictionary(g => g.Key, g => g.Select(i => i.Key));
+        }
```
mikeminutillo: Given that QueueLength is reporting EndpointInstanceIds, do we need this to return a dictionary or can it just return a list of all the instance ids?

tmasternak: I think list should be fine

tmasternak:
@mikeminutillo I say merge let's merge

mikeminutillo:
Done

PR #41, opened 2017-07-26
Refactoring IntervalStores

tmasternak:
This is an effort to apply SRP principles to the current structure of our solution. Here are the main points:

* `IntervalStore` is responsible only for storing and returning intervals - nothing more
*  `TimingAggregator` has been changed to static class that operates on intervals and belongs to the endpoints list part of API,
* `MonitoredEndpointsModule` has an opinion on what stores it wants (this is driven by concrete UI pages) however it uses generic method to aggregate stored data into endpoint or instance specific types
 * Concrete stores are just marker types,
 * Bunch of nameology


src/ServiceControl.Monitoring.Tests/QueueLength/QueueLengthStoreTests.cs
```diff
@@ -194,11 +194,10 @@ public void If_send_report_is_received_value_snapshot_is_taken_for_endpoints_wit
...
             consumer.SnapshotCurrentQueueLengthEstimations(now);
 
-            var lengths = new MonitoredEndpoint { Name = "ReceiverA" };
-            consumer.FillIn(new[] { lengths }, now);
+            var lengths = consumer.GetIntervals(now);
 
-            Assert.IsNotNull(lengths.QueueLength);
-            Assert.AreEqual(3 - 1, lengths.QueueLength.Points.Last());
+            Assert.AreEqual(1, lengths.Length);
+            Assert.AreEqual(3 - 1, lengths[0].Intervals.First().TotalValue);
```
Scooletz: I'd extract `3` and `1` to consts to make `1` different from the `1` in the line 199.

src/ServiceControl.Monitoring/Infrastructure/EndpointRegistry.cs
```diff
@@ -6,18 +6,23 @@
...
         {
-            endpointInstances.Add(endpointInstanceId);    
+            endpointInstances.TryAdd(endpointInstanceId, true);    
         }
 
-        public IDictionary<string, IEnumerable<EndpointInstanceId>> AllEndpoints()
+        public IDictionary<string, IEnumerable<EndpointInstanceId>> GetAllEndpoints()
         {
-            return endpointInstances.ToArray()
+            return endpointInstances.Keys.ToArray()
```
Scooletz: `endpointInstances.Select(kvp=>kvp.Keys) ` would not issue the lock (the non-locking concurrent iterator). Should we consider it in here?

src/ServiceControl.Monitoring/Infrastructure/EndpointRegistry.cs
```diff
@@ -6,18 +6,23 @@
...
                 .GroupBy(i => i.EndpointName)
                 .ToDictionary(g => g.Key, g => g.Select(i => i));
         }
+
+        public EndpointInstanceId[] GetEndpointInstances(string endpointName)
+        {
+            return GetAllEndpoints()[endpointName].ToArray();
```
Scooletz: So we get all the endpoints to select just one? Could we do better?

PR #42, opened 2017-07-27
Performance tests of core Monitoring components

Scooletz:
This PR introduces performance tests that are meant to stress the write and the read side of `SC.Monitoring`. It does it following way by:
- simulating a run of a number of scaled-up endpoints, reporting every N milliseconds
- reading the data at the same time
- measuring the read and the write side with `HdrHistogram`.

### How it works
The tests skips the transport layer and use a direct access to the components which are normally executed by handlers to register values. On the other side, it uses the Nancy module as a regular object, without calling it via `http`. This provides a test that measures the write/read side of the `SC.Monitoring` itself, **excluding**:
- the transport overhead
- Nancy/http overhead of delivering value

As these two above are well known, in my opinion it's still valuable to deliver boundaries for the execution of the core `SC.Monitoring` components in "almost end-to-end" test suite.

### Results
Initial tests for a request under load of thousand of instances:

|Side| Queue length| Critical time/Processing time/Retries |
|---|---|---|
|Writes| 0.8 ms  |0.03 ms|
|Reads| 0.5 ms  | 14 ms|


Scooletz:
@tmasternak Is this something we'd like to follow?

src/ServiceControl.Monitoring.Tests/QueueLength/QueueLengthStoreTests.cs
```diff
@@ -202,7 +202,7 @@ public void If_send_report_is_received_value_snapshot_is_taken_for_endpoints_wit
 
         }
 
-        static JObject BuildMessage(Counter[] sendCounters = null, Gauge[] receiveGauges = null)
+        public static JObject BuildMessage(Counter[] sendCounters = null, Gauge[] receiveGauges = null)
```
tmasternak: Can we extract that to separate builder classes?

src/ServiceControl.Monitoring.Tests/PerformanceTests.cs
```diff
@@ -0,0 +1,236 @@
...
+                var receivedSessions = Enumerable.Range(0, numberOfReceivedSessionsPerEndpoint)
+                    .Select(_ => random.Next(sessionCount))
+                    .Select(si => allSessions[si]).ToArray();
+
+                endpointReporters[i] = BuildQueueLengthReporter(allEndpoints[i], receivedSessions, sentSessions, queueLengthStore, source);
+            }
+
+            var snapshoter = Task.Run(async () =>
+            {
+                for (var i = 0; i < 200; i++)
```
tmasternak: Do we want to extract that to const?

src/ServiceControl.Monitoring.Tests/PerformanceTests.cs
```diff
@@ -0,0 +1,236 @@
...
+                histogram.RecordValue(elapsed);
+            }
+
+            source.Cancel();
+            await Task.WhenAll(endpointReporters).ConfigureAwait(false);
+
+            var reportFinalHistogram = MergeHistograms(endpointReporters);
+
+            PrintHistogram("Querying", histogram);
+            PrintHistogram("Reporters", reportFinalHistogram);
```
tmasternak: How about we add asserts with some baseline numbers?

mikeminutillo:
Awesome. Can we try running this against ASB/ASQ. The ASQ tests for SC run twice as long on average for SC (although to be fair, that's NSB v5)

Scooletz:
@mikeminutillo I'll update a description of this as it does not include transport. If that was your understanding then I documented it poorly. I'll ping you as soon as I make it better.

Scooletz:
Ping @mikeminutillo I've updated the description to address the doubts raised by your last comment. Could you take a look whether it's more descriptive? Meanwhile, I'll make adjustments raised by @tmasternak

Scooletz:
The build is getting a bit lengthy because of the perf tests runs. Are you ok with it @Particular/servicecontrol-maintainers ?
tests_runs

WilliamBZA:
:+1:

Scooletz:
I'm dismissing @tmasternak review. I've addressed all the points.

Scooletz:
Ping @Particular/servicecontrol-maintainers it's ready for the review

src/ServiceControl.Monitoring/Hosting/Commands/RunCommand.cs
```diff
@@ -52,7 +52,7 @@ class OnConsoleCancel
         {
             Action action;
 
-            private OnConsoleCancel(Action action)
+            OnConsoleCancel(Action action)
```
mikeminutillo: Not that I object to this change, but why is it in here?

src/ServiceControl.Monitoring/Infrastructure/EndpointRegistry.cs
```diff
@@ -1,30 +1,33 @@
...
 
         public void Record(EndpointInstanceId endpointInstanceId)
         {
-            endpointInstances.TryAdd(endpointInstanceId, true);    
+            lock (@lock)
+            {
+                endpointInstances.Add(endpointInstanceId);
+                lookup = endpointInstances.ToArray()
+                    .GroupBy(instance => instance.EndpointName)
+                    .ToDictionary(g => g.Key, g => (IEnumerable<EndpointInstanceId>)g.Select(i => i).ToArray());
```
mikeminutillo: Why do we convert to an array but cast as an IEnumerable? Should this be an IReadOnlyCollection maybe?

src/ServiceControl.Monitoring/QueueLength/TagExtensions.cs
```diff
@@ -20,10 +20,13 @@ public static bool TryGetTagValue(this IEnumerable<string> tags, string key, out
...
-            value = tags
-                .Where(t => t.StartsWith(prefix, StringComparison.OrdinalIgnoreCase))
-                .Select(x => x.Substring(key.Length + 1).Trim())
-                .FirstOrDefault();
+            value = tags.FirstOrDefault(t => t.StartsWith(prefix, StringComparison.OrdinalIgnoreCase));
+
+            // ReSharper disable once UseNullPropagation
+            if (value != null)
+            {
+                value = value.Substring(key.Length + 1).Trim();
```
mikeminutillo: Could this just be `prefix.Length`?

src/ServiceControl.Monitoring.Tests/PerformanceTests.cs
```diff
@@ -0,0 +1,245 @@
...
+                endpointReporters[i] = BuildQueueLengthReporter(allEndpoints[i], receivedSessions, sentSessions, queueLengthStore, source);
+            }
+
+            var snapshoter = Task.Run(async () =>
+            {
+                for (var i = 0; i < totalQueryCount; i++)
+                {
+                    queueLengthStore.SnapshotCurrentQueueLengthEstimations(DateTime.UtcNow);
+
+                    await Task.Delay(TimeSpan.FromMilliseconds(10));
```
mikeminutillo: Is this 10s the same as the below 10s? Should this be a variable/const somewhere?

src/ServiceControl.Monitoring.Tests/PerformanceTests.cs
```diff
@@ -0,0 +1,245 @@
...
+                await GetMonitoredEndpoints().ConfigureAwait(false);
+                var elapsed = Stopwatch.GetTimestamp() - start;
+                histogram.RecordValue(elapsed);
+            }
+
+            source.Cancel();
+            await Task.WhenAll(reporters).ConfigureAwait(false);
+
+            MergeHistograms(reporters);
+            var reportFinalHistogram = MergeHistograms(reporters);
```
mikeminutillo: Why the double merge?

mikeminutillo:
> I've updated the description to address the doubts raised by your last comment. Could you take a look whether it's more descriptive?

:+1: So we probably still want to run some kind of transport saturation test to see if we can drive the core this fast across the different transports.

src/ServiceControl.Monitoring/Hosting/Commands/RunCommand.cs
Scooletz: This is my `TeamCity` kungfu failing me when checking for inspection errors. I can undo it if needed.

src/ServiceControl.Monitoring/Infrastructure/EndpointRegistry.cs
Scooletz: My reasoning behind it:
- `.ToArray` to materialize it
- cast to `IEnumerable` to make it immutable (at least via published interface)

src/ServiceControl.Monitoring/QueueLength/TagExtensions.cs
Scooletz: Oh.... yes 😊

src/ServiceControl.Monitoring.Tests/PerformanceTests.cs
Scooletz: Will deduplicate it 👍

src/ServiceControl.Monitoring.Tests/PerformanceTests.cs
Scooletz: This is an a value bigger than a current snaphotting ratio. My aim was to put more pressure on the snapshotting. I can extract it to a local var. I see no way/reason to make it a parameter for the test itself.

Scooletz:
> 👍 So we probably still want to run some kind of transport saturation test to see if we can drive the core this fast across the different transports.

Could you explain your reasoning a bit more? In my opinion, taking into consideration the encoding we use now, the way we deserialize, it's unlikely to saturate transports. On the other hand, maybe it'd be good to see one real number. I think that running this test all the time would be unneeded though. In my opinion we'll be bound by the internal structures, not the transport. 

Should this be a separate issue or should it be included in this PR?

Scooletz:
@mikeminutillo Sorry but I rebased this PR on top of `develop` and pushed force it. It was failing because of the introduction of `VariableHistoryIntervalStore`.

mikeminutillo:
@Scooletz I think we should do it as another PR, rather than here. 

I agree with you, it's not a test that we should run all of the time and the work that's been done to make the wire format performant makes this a small concern. I want to ensure we don't get to the end of 3 months of development _and then_ discover that we're sending too much data over the wire. Especially for customers where message count is a limited resource. It's not just message size that I'm thinking of, but message count as well. For customers in cloud environments they may have quotas and we may want a dedicated namespace for monitoring or they may be paying by message in which case we want to batch up more info per message. The earlier we can find these things out, the better.

Scooletz:
@mikeminutillo Could you give this PR 👍 or 👎 ? It has not been approved nor rejected yet.

mikeminutillo:
:+1: just waiting for build now and I'll merge

PR #43, opened 2017-07-28
Introducing variable history interval store

tmasternak:
## Description
This PR introduces `VariableHistoryIntervalStoreTests` that aggregates `IntervalStores` - one per each supported historical period. This PR also introduces changes to api module enabling queries for specific time periods.


src/ServiceControl.Monitoring/Infrastructure/IntervalsStore.cs
```diff
@@ -172,9 +181,14 @@ public class TimeInterval
...
+        public IntervalsStore(TimeSpan intervalSize, int numberOfIntervals)
+        {
+            this.intervalSize = intervalSize;
+            this.numberOfIntervals = numberOfIntervals;
+        }
 
-        static TimeSpan IntervalSize = TimeSpan.FromSeconds(15);
+        /// Number of 15s intervals in 5 minutes
+        int numberOfIntervals = 4 * 5;
+        TimeSpan intervalSize = TimeSpan.FromSeconds(15);
```
mikeminutillo: With no default constructor, these values will never be used.

src/ServiceControl.Monitoring/Infrastructure/HistoryPeriod.cs
```diff
@@ -0,0 +1,65 @@
...
+        public override int GetHashCode()
+        {
+            return Value.GetHashCode();
+        }
+
+        static readonly Dictionary<TimeSpan, HistoryPeriod> all;
+
+        public static IReadOnlyCollection<HistoryPeriod> All;
+    }
+}
```
mikeminutillo: Given the small count of these, is the lookup dictionary really necessary? You could just `FirstOrDefault()` the readonly collection and throw on a null response.

src/ServiceControl.Monitoring/Infrastructure/VariableHistoryIntervalStore.cs
```diff
@@ -0,0 +1,45 @@
...
+            {
+                return store.GetIntervals(now);
+            }
+
+            throw new Exception("Unsupported history size.");
+        }
+
+        public void Store(EndpointInstanceId instanceId, RawMessage.Entry[] entries)
+        {
+            foreach (var kvHistory in histories.ToArray())
```
mikeminutillo: Is this `ToArray()` there to lock the collection before storing? Will it do that? We don't actually need to as we never mess with the `histories` collection after we create it. Can we just use `histories.Values` instead?

src/ServiceControl.Monitoring/Infrastructure/VariableHistoryIntervalStore.cs
```diff
@@ -0,0 +1,45 @@
...
+            {
+                kvHistory.Value.Store(instanceId, entries);
+            }
+        }
+
+        const int IntervalsPerStore = 20;
```
mikeminutillo: Is this based on the UI? Do we currently show 20 samples?

src/ServiceControl.Monitoring/Http/Diagrams/DiagramApiModule.cs
```diff
@@ -18,36 +18,43 @@ public class MonitoredEndpointsModule : ApiModule
         /// </summary>
         public MonitoredEndpointsModule(EndpointRegistry endpointRegistry, CriticalTimeStore criticalTimeStore, ProcessingTimeStore processingTimeStore, RetriesStore retriesStore, QueueLengthStore queueLengthStore)
         {
-            Get["/monitored-endpoints"] = x =>
+            const int DefaultHistory = 5;
+
+            Get["/monitored-endpoints"] = parameters =>
             {
                 var endpoints = GetMonitoredEndpoints(endpointRegistry);
+                var period = HistoryPeriod.FromMinutes((int?)Request.Query["history"] ?? DefaultHistory);
```
mikeminutillo: does the parameters collection also cover `Request.Query`? I think this could be `(int?)parameters.history) ?? DefaultHistory`.

src/ServiceControl.Monitoring/Infrastructure/IntervalsStore.cs
tmasternak: Correct. My bad

src/ServiceControl.Monitoring/Infrastructure/HistoryPeriod.cs
tmasternak: Good point

src/ServiceControl.Monitoring/Infrastructure/VariableHistoryIntervalStore.cs
tmasternak: I think you made a good Point. The whole Thing could be just a Dictionary

src/ServiceControl.Monitoring/Infrastructure/VariableHistoryIntervalStore.cs
tmasternak: Yes

src/ServiceControl.Monitoring/Http/Diagrams/DiagramApiModule.cs
tmasternak: Did not know that. I like parameters better

WilliamBZA: parameters is not populated by query string parameters.

src/ServiceControl.Monitoring.Tests/Infrastructure/EntriesBuilder.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using Messaging;
+    using NUnit.Framework;
+
+    public static class EntriesBuilder
+    {
+        public static RawMessage.Entry[] Build(Dictionary<DateTime, long> measurements)
```
Scooletz: If this wasn't for tests, I'd argue that an array of `{DateTime,long}` tuples would be much better. It's fine for test though.

src/ServiceControl.Monitoring/Infrastructure/VariableHistoryIntervalStore.cs
```diff
@@ -0,0 +1,45 @@
+﻿namespace ServiceControl.Monitoring.Infrastructure
+{
+    using System;
+    using System.Collections.Concurrent;
+    using Messaging;
+
+    public abstract class VariableHistoryIntervalStore
+    {
+        ConcurrentDictionary<HistoryPeriod, IntervalsStore> histories;
```
Scooletz: We've got `HistoryPeriod.All`, so this could be made a regular `Dictionary`. The stores underneath are mutable and thread safe so there's no gain in using a `ConcurrentDict` in here.

WilliamBZA:
Thanks @Scooletz, I've removed the concurrent dictionary.

src/ServiceControl.Monitoring/Http/Diagrams/DiagramApiModule.cs
mikeminutillo: `throw new InsufficientMagicException()`

PR #47, opened 2017-08-01
Throughput in msg/s

Scooletz:
Connects to https://github.com/Particular/LaunchWave.DevOpsAdvancedMonitoring/issues/143

This PR introduced Throughput in `messages/s` available for both: endpoints and their instances. 

This is done by:
- introducing a new aggregating function `IntervalsAggregator.AggregateTotalMeasurementsPerSecond`
- applying it on any store that measures per message statistic (`processingTimeStore` has been selected)
- dividing any obtained result by the number of seconds in one interval of the specific `HistoryPeriod` that the report is run for


Scooletz:
Good point @mikeminutillo As far as I see it might be not reporter if the pipeline blows out:

https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Pipeline/MainPipelineExecutor.cs#L34

My questions is: should I take it into a consideration?

WilliamBZA:
@mikeminutillo Processing time should NOT be reported for failed messages. The processing time metric is how long it took to successfully process the message, excluding retries etc.

Scooletz:
@WilliamBZA Does it make this calculations wrong or right? What's the `throughput`? `The number of successfully processed messages per seconds` or `The number of retrieved messages per seconds` ?

WilliamBZA:
I think that Throughput should also be based on successful messages, so I think we're good here.

WilliamBZA:
By that I mean that I don't think it makes sense to say that an endpoint is processing 300 messages a second, but those are all trapped in FLR+SLR and the actual number of successful messages is more like 5 messages a second.

src/ServiceControl.Monitoring/Http/Diagrams/DiagramApiModule.cs
```diff
@@ -38,18 +39,24 @@ public MonitoredEndpointsModule(EndpointRegistry endpointRegistry, CriticalTimeS
...
                 FillInInstanceData(instances, retriesStore, period, (e, v) => e.Retries = v, IntervalsAggregator.AggregateRetries);
+                FillInInstanceData(instances, processingTimeStore, period, (e, v) => e.Throughput = v, intervals => IntervalsAggregator.AggregateTotalMeasurementsPerSecond(intervals, period));
 
                 return Negotiate.WithModel(instances);
             };
         }
 
-        static void FillInEndpointData(MonitoredEndpoint[] endpoints, 
-            VariableHistoryIntervalStore store, 
+        HistoryPeriod ExtractHistoryPeriod()
```
WilliamBZA: ❤️

PR #48, opened 2017-08-02
Throughput calculation fixup

Scooletz:
Throughput should be calculated in a following way:
- Average: average over all reported intervals, 0 when no reports
- Points: should provide a summarized value per endpoint/instance

to make it `msg/s` the result value in both cases is divided by number of seconds. This makes it coheren with Retries aggregation beside the final division by seconds.


Scooletz:
Ping @Particular/servicecontrol-maintainers Could you review and merge it?

PR #49, opened 2017-08-02
Monitoring endpoint instance activity

Scooletz:
Connects to https://github.com/Particular/LaunchWave.DevOpsAdvancedMonitoring/issues/147

This PR introduces the monitoring of endpoints' instances' activity. 

When no message is sent for a specific period of time, an instance is considered stale and this state is reported via `MonitoredEndpointInstance.IsStale` property. 
No staleness is calculated for the `MonitoredEndpoint` as there's no single value that can describe whether or not an endpoint's data, as a whole, are stale or not.

### DIFF
--- a/src/ServiceControl.Monitoring/Infrastructure/EndpointInstanceActivityTracker.cs
+++ b/src/ServiceControl.Monitoring/Infrastructure/EndpointInstanceActivityTracker.cs
@@ -1,0 +1,28 @@
+﻿namespace ServiceControl.Monitoring.Infrastructure
+{
+    using System;
+    using System.Collections.Concurrent;
+
+    public class EndpointInstanceActivityTracker
+    {
+        ConcurrentDictionary<EndpointInstanceId, DateTime> endpointsInstances = new ConcurrentDictionary<EndpointInstanceId, DateTime>();
+        public static readonly TimeSpan StalenessThreshold = TimeSpan.FromMinutes(5);
+
+        public void Record(EndpointInstanceId instanceId, DateTime utcNow)
+        {
+            endpointsInstances.AddOrUpdate(instanceId, utcNow, (_, __) => utcNow);
+        }
+
+        public bool IsStale(EndpointInstanceId endpointInstance)
+        {
+            DateTime lastActivityTime;
+            if (endpointsInstances.TryGetValue(endpointInstance, out lastActivityTime))
+            {
+                var timeSpan = DateTime.UtcNow - lastActivityTime;
+                return timeSpan > StalenessThreshold;
+            }
+
+            return false;
+        }
+    }
+}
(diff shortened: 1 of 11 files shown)


Scooletz:
@WilliamBZA @WojcikMike Could you take a look at it? Is this the way we'd like to report the staleness of data?

src/ServiceControl.Monitoring/Infrastructure/EndpointTracker.cs
```diff
@@ -1,15 +1,17 @@
...
+    using System;
     using System.Threading.Tasks;
     using Messaging;
     using NServiceBus;
     using NServiceBus.Metrics;
 
     public class EndpointTracker : IHandleMessages<MetricReport>, IHandleMessages<LongValueOccurrences>, IHandleMessages<Occurrences>
     {
-        public EndpointTracker(EndpointRegistry endpointRegistry)
+        public EndpointTracker(EndpointRegistry endpointRegistry, EndpointInstanceActivityTracker activityTracker)
```
WilliamBZA: Is the idea that any future metrics that get added will come through here as either a `LongValueOccurrences`, `MetricReport`, or `Occurrences`?

Or if it's a new type of metric report/occurrence that the handler will live in this class as well?

src/ServiceControl.Monitoring/Infrastructure/EndpointInstanceActivityTracker.cs
```diff
@@ -0,0 +1,28 @@
+﻿namespace ServiceControl.Monitoring.Infrastructure
+{
+    using System;
+    using System.Collections.Concurrent;
+
+    public class EndpointInstanceActivityTracker
+    {
+        ConcurrentDictionary<EndpointInstanceId, DateTime> endpointsInstances = new ConcurrentDictionary<EndpointInstanceId, DateTime>();
+        public static readonly TimeSpan StalenessThreshold = TimeSpan.FromMinutes(5);
```
WilliamBZA: We must just load this time from the Settings class. We'd have to introduce a new setting for it too.

Scooletz: Any link to how we do it now? Any helper classes etc?

src/ServiceControl.Monitoring/Infrastructure/EndpointTracker.cs
Scooletz: I thought that there was one handler for all the messages but then I saw this that handles them explicitly. My assumption was to do not change everything and just attach to the thing that registers endpoints' instances in a registry. This might be a subject to change in the future (but then again, it will cover the registry and the activity tracker at the same time).

Does it answer you question?

src/ServiceControl.Monitoring/Infrastructure/EndpointInstanceActivityTracker.cs
WilliamBZA: I've pushed the code to do it

src/ServiceControl.Monitoring/Infrastructure/EndpointTracker.cs
WilliamBZA: :+1:

WojcikMike:
LGTM

WilliamBZA:
Gah, pushed a fix for the tests. Once build is green I'll merge

PR #54, opened 2017-08-30
Metrics breakdown by message type

tmasternak:
This PR introduces metric values broken down by message type. This feature was introduced by making following changes:
 * Introducing deserializer for `TaggedLongValueOccurrence` monitoring messages,
 * Making `VariableHistoryIntervalStore` and `IntervalStore` generic of `T`. Where `T` is a type by which the store is breaking the data. Currently there are two values for `T` - `EndpointInstanceId` and `EndpointMessageType`. That allows storing measurement intervals by instance and/or message type
  * Adding `TaggedLongValueOccurenceHandler` that stores tagged measurements in `ProcessingTimeStore` and `CriticalTimeStore`,

On the query side this PR makes the query logic be more declarative and OCP (inspired by #39 by @mikeminutillo):
  * Stores are implementing `IProvideBreakdownBy<T>`
  * `DiagramsApi` has constructor dependency on implementations of those
  * `DiagramsApi` is using `MonitoredMetric` classes to specify what metric is to be returned by the api, under what name and what aggregation function should be used.


src/ServiceControl.Monitoring.Tests/Infrastructure/IntervalStoreTests.cs
```diff
@@ -162,9 +162,32 @@ public void Intervals_are_returned_in_descending_order()
...
         }
 
-        IntervalsStore AnyStore()
+        IntervalsStore<BreakdownId> AnyStore()
         {
-            return new IntervalsStore(TimeSpan.FromSeconds(5), 127);
+            return new IntervalsStore<BreakdownId>(TimeSpan.FromSeconds(5), 127);
+        }
+
+        class BreakdownId
```
Scooletz: There's no generic requirement for the type of breakdown. Why not to use some primitive, like `int`?

src/ServiceControl.Monitoring.Tests/Infrastructure/VariableHistoryIntervalStoreTests.cs
```diff
@@ -39,9 +39,28 @@ public void Store_updates_all_supported_historical_periods()
                 Assert.AreEqual(1L, intervals[0].TotalMeasurements);
             }
         }
-    }
 
-    public class SomeStore : VariableHistoryIntervalStore
-    {
+        class BreakdownId
```
Scooletz: There's no generic requirement for the type of breakdown. Why not to use some primitive, like `int`?

src/ServiceControl.Monitoring.Tests/PerformanceTests.cs
```diff
@@ -244,5 +259,55 @@ static void Report(string name, LongHistogram histogram, TimeSpan? maximumMean)
                 Assert.LessOrEqual(actualMean, max, $"The actual mean for {name} was '{actualMean}' and was bigger than maximum allowed mean '{max}'.");
             }
         }
+
+        interface IStoreMetric
```
Scooletz: This could be replaced by a lambda `Action<RawMessage.Entry[], EndpointInstanceId instanceId> store` in `StartReporter` with no need of creating wrappers and the interface itself.

src/ServiceControl.Monitoring/Http/Diagrams/DiagramApiModule.cs
```diff
@@ -15,78 +15,118 @@ public class MonitoredEndpointsModule : ApiModule
...
-               
-                FillInEndpointData(endpoints, criticalTimeStore, period, (e, v) => e.CriticalTime = v, IntervalsAggregator.AggregateTimings);
-                FillInEndpointData(endpoints, processingTimeStore, period, (e, v) => e.ProcessingTime = v, IntervalsAggregator.AggregateTimings);
-                FillInEndpointData(endpoints, retriesStore, period, (e, v) => e.Retries = v, IntervalsAggregator.AggregateRetries);
-                FillInEndpointData(endpoints, queueLengthStore, period, (e, v) => e.QueueLength = v, IntervalsAggregator.AggregateQueueLength);
-                FillInEndpointData(endpoints, processingTimeStore, period, (e, v) => e.Throughput = v, intervals => IntervalsAggregator.AggregateTotalMeasurementsPerSecond (intervals, period));
+
+                foreach (var metric in instanceMetric)
+                {
+                    var store = metricByInstance.Single(s => s.GetType() == metric.StoreType);
```
Scooletz: This should be a dictonary initialized as a field in ctor and lookup should be done by indexer `[metric.StoreType]`

src/ServiceControl.Monitoring/Http/Diagrams/DiagramApiModule.cs
```diff
@@ -15,78 +15,118 @@ public class MonitoredEndpointsModule : ApiModule
...
                 var instances = GetMonitoredEndpointInstances(endpointRegistry, endpointName, activityTracker);
-                var period = ExtractHistoryPeriod();
 
-                FillInInstanceData(instances, criticalTimeStore, period, (e, v) => e.CriticalTime = v, IntervalsAggregator.AggregateTimings);
-                FillInInstanceData(instances, processingTimeStore, period, (e, v) => e.ProcessingTime = v, IntervalsAggregator.AggregateTimings);
-                FillInInstanceData(instances, retriesStore, period, (e, v) => e.Retries = v, IntervalsAggregator.AggregateRetries);
-                FillInInstanceData(instances, processingTimeStore, period, (e, v) => e.Throughput = v, intervals => IntervalsAggregator.AggregateTotalMeasurementsPerSecond(intervals, period));
+                foreach (var metric in instanceMetric)
+                {
+                    var store = metricByInstance.Single(s => s.GetType() == metric.StoreType);
```
Scooletz: This should be a dictionary initialized as a field in ctor and lookup should be done by indexer [metric.StoreType]

src/ServiceControl.Monitoring/Http/Diagrams/DiagramApiModule.cs
```diff
@@ -15,78 +15,118 @@ public class MonitoredEndpointsModule : ApiModule
...
-        {
-            var intervals = store.GetIntervals(period, DateTime.UtcNow).ToLookup(k => k.Id.EndpointName);
+                var messageTypes = GetMonitoredMessageTypes(metricByMessageType, period, endpointName);
 
-            foreach (var endpoint in endpoints)
-            {
-                var values = aggregate(intervals[endpoint.Name].ToList());
+                foreach (var metric in messageTypeMetric)
+                {
+                    var store = metricByMessageType.Single(s => s.GetType() == metric.StoreType);
```
Scooletz: This should be a dictionary initialized as a field in ctor and lookup should be done by indexer [metric.StoreType]

src/ServiceControl.Monitoring/Http/Diagrams/DiagramApiModule.cs
```diff
@@ -15,78 +15,118 @@ public class MonitoredEndpointsModule : ApiModule
...
+                    var intervals = store.GetIntervals(period, DateTime.UtcNow).ToLookup(k => k.Id);
 
-                setter(endpoint, values);
-            }
-        }
+                    foreach (var messageType in messageTypes)
+                    {
+                        var values = metric.Aggregate(intervals[new EndpointMessageType(endpointName, messageType.MessageType)].ToList(), period);
```
Scooletz: It looks quite complex: double loop, some `ToList()`. Should we consider a benchmark for it?

src/ServiceControl.Monitoring/Http/Diagrams/DiagramApiModule.cs
```diff
@@ -15,78 +15,118 @@ public class MonitoredEndpointsModule : ApiModule
...
-        {
-            var intervals = store.GetIntervals(period, DateTime.UtcNow).ToLookup(k => k.Id.InstanceId);
+                        messageType.Metrics.Add(metric.ReturnName, values);
+                    }
+                }
 
-            foreach (var instance in instances)
-            {
-                var values = aggregate(intervals[instance.Id].ToList());
+                var data = new MonitoredEndpointDetails
```
Scooletz: Object for the response FTW!

src/ServiceControl.Monitoring/Http/Diagrams/DiagramApiModule.cs
```diff
@@ -107,8 +147,21 @@ static MonitoredEndpoint[] GetMonitoredEndpoints(EndpointRegistry endpointRegist
...
                     Name = endpoint.Key,
                     EndpointInstanceIds = endpoint.Value.Select(i => i.InstanceId).ToArray(),
-                    IsStale = endpoint.Value.Any(i => activityTracker.IsStale(i))
+                    IsStale = endpoint.Value.Any(activityTracker.IsStale)
                 }).ToArray();
         }
+
+        static MonitoredEndpointMessageType[] GetMonitoredMessageTypes(IEnumerable<IProvideBreakdownBy<EndpointMessageType>> stores, HistoryPeriod period, string endpointName)
+        {
+            var messageTypes = stores.SelectMany(s => s.GetIntervals(period, DateTime.UtcNow))
```
Scooletz: This looks really heavy. Any tests for 100 endpoints with 100 message types each?

src/ServiceControl.Monitoring/Http/Diagrams/IntervalsAggregator.cs
```diff
@@ -37,11 +37,11 @@ internal static MonitoredEndpointValues AggregateRetries(List<IntervalsStore.End
             };
         }
 
-        internal static MonitoredEndpointValues AggregateQueueLength(List<IntervalsStore.EndpointInstanceIntervals> intervals)
+        internal static MonitoredValues AggregateQueueLength<T>(List<IntervalsStore<T>.IntervalsBreakdown> intervals, HistoryPeriod period)
```
Scooletz: Just thinking now. We could use arrays instead of lists in all these methods, correct?

tmasternak: Yes this is correct. That being this in worst case this collection holds as many items as there are instances, message-types per logical endpoint - so I'm not sure how much we can gain here. 

I propose to wait until we have performance tests for query side.

src/ServiceControl.Monitoring/Http/Diagrams/DiagramApiModule.cs
tmasternak: I'll work on performance test and after that we can re-evaluate.

src/ServiceControl.Monitoring/Http/Diagrams/DiagramApiModule.cs
tmasternak: see https://github.com/Particular/ServiceControl.Monitoring/pull/54#discussion_r136049553

tmasternak:
I need to update `NServiceBus.Metrics` reference after https://github.com/Particular/NServiceBus.Metrics/pull/54 is merged.

tmasternak:
@mikeminutillo it would be great to get your review. It is still `WIP` but that's mainly due to missing performance test for query side.

src/ServiceControl.Monitoring/QueueLength/QueueLengthStore.cs
```diff
@@ -7,13 +7,20 @@ namespace ServiceControl.Monitoring.QueueLength
...
 
-    public class QueueLengthStore : VariableHistoryIntervalStore
+    public class QueueLengthStore : IProvideBreakdownBy<EndpointInstanceId>
     {
         public QueueLengthStore(IQueueLengthCalculator calculator)
         {
             this.calculator = calculator;
         }
 
+        VariableHistoryIntervalStore<EndpointInstanceId> byInstance { get; } = new VariableHistoryIntervalStore<EndpointInstanceId>();
```
mikeminutillo: Why a property and not just a readonly field?

src/ServiceControl.Monitoring/Timings/LongValueOccurrenceHandler.cs
```diff
@@ -23,16 +23,16 @@ public Task Handle(LongValueOccurrences message, IMessageHandlerContext context)
         {
             try
             {
-                var endpointInstanceId = EndpointInstanceId.From(context.MessageHeaders);
+                var instanceId = EndpointInstanceId.From(context.MessageHeaders);
```
mikeminutillo: Would it make sense to collapse this with `TaggedLongValueOccurrenceHandler` into a single `TimingsHandler`?

src/ServiceControl.Monitoring/Infrastructure/EndpointMessageType.cs
```diff
@@ -0,0 +1,40 @@
+﻿namespace ServiceControl.Monitoring.Infrastructure
+{
+    public class EndpointMessageType
+    {
+        public EndpointMessageType(string endpointName, string messageType)
```
mikeminutillo: This might get more complicated in the real world. If I have two messages of the same name but different assembly versions, should they be combined or shown separately? 

I think this is good enough for now though and it's really nicely abstracted so we can mess with it later.

src/ServiceControl.Monitoring/Retries/RetriesStore.cs
```diff
@@ -1,7 +1,28 @@
...
+    using Messaging;
+
+    public class RetriesStore : IProvideBreakdownBy<EndpointInstanceId>, IProvideBreakdownBy<EndpointMessageType>
     {
+        VariableHistoryIntervalStore<EndpointInstanceId> byInstance { get; } = new VariableHistoryIntervalStore<EndpointInstanceId>();
+        VariableHistoryIntervalStore<EndpointMessageType> byMessageType { get; } = new VariableHistoryIntervalStore<EndpointMessageType>();
+
+        public IntervalsStore<EndpointInstanceId>.IntervalsBreakdown[] GetIntervals(HistoryPeriod period, DateTime now)
+        {
+            return byInstance.GetIntervals(period, now);
```
mikeminutillo: Given that this implementation is the same for every implementation of `IProvideBreakDownBy<>` we could consider just exposing the property on the interface. Each _store_ would need to explicitly implement each of the interfaces but that becomes one-line. For every downstream that needs to consume it you would do

```
stores.OfType<TBreakdown>().Select(x => x.Breakdown.GetIntervals(period, now));
```

Just a thought.

mikeminutillo: It does get messier to access them internally though

src/ServiceControl.Monitoring/Retries/RetriesStore.cs
```diff
@@ -1,7 +1,28 @@
...
+
+        IntervalsStore<EndpointMessageType>.IntervalsBreakdown[] IProvideBreakdownBy<EndpointMessageType>.GetIntervals(HistoryPeriod period, DateTime now)
+        {
+            return byMessageType.GetIntervals(period, now);
+        }
+
+        public void Store(RawMessage.Entry[] entries, EndpointInstanceId instanceId, EndpointMessageType messageType)
+        {
+            byInstance.Store(instanceId, entries);
+            byMessageType.Store(messageType, entries);
```
mikeminutillo: Is this future proofing? At the moment we don't tag retries right?

src/ServiceControl.Monitoring/Http/Diagrams/DiagramApiModule.cs
```diff
@@ -15,78 +15,119 @@ public class MonitoredEndpointsModule : ApiModule
...
+                }
 
-        static void FillInEndpointData(MonitoredEndpoint[] endpoints,
-            VariableHistoryIntervalStore store,
-            HistoryPeriod period,
-            Action<MonitoredEndpoint, MonitoredEndpointValues> setter,
-            Func<List<IntervalsStore.EndpointInstanceIntervals>, MonitoredEndpointValues> aggregate)
-        {
-            var intervals = store.GetIntervals(period, DateTime.UtcNow).ToLookup(k => k.Id.EndpointName);
+                var messageTypes = GetMonitoredMessageTypes(metricByMessageType, period, endpointName);
```
mikeminutillo: I feel like this this should be a separate HTTP call. It feels very orthogonal to the code above

src/ServiceControl.Monitoring/Http/Diagrams/DiagramApiModule.cs
```diff
@@ -107,8 +148,21 @@ static MonitoredEndpoint[] GetMonitoredEndpoints(EndpointRegistry endpointRegist
                 {
                     Name = endpoint.Key,
                     EndpointInstanceIds = endpoint.Value.Select(i => i.InstanceId).ToArray(),
-                    IsStale = endpoint.Value.Any(i => activityTracker.IsStale(i))
+                    IsStale = endpoint.Value.Any(activityTracker.IsStale)
                 }).ToArray();
         }
+
+        static MonitoredEndpointMessageType[] GetMonitoredMessageTypes(IEnumerable<IProvideBreakdownBy<EndpointMessageType>> stores, HistoryPeriod period, string endpointName)
```
mikeminutillo: Maybe we should a monitored message type registry in the same way that we have an endpoint registry?

src/ServiceControl.Monitoring/Retries/RetriesStore.cs
Scooletz: Actually we do. We tag any `duration` or `signal`. This means that every thing we measure now falls into one of these buckets.

src/ServiceControl.Monitoring/Infrastructure/EndpointMessageType.cs
Scooletz: IMO this should be done either in SC.Monitoring or maybe client itself (it's an easy aggregation).

tmasternak: As a side note: currently the string holds the type full name so we do have assembly, version and public key and can parse that in SC.Monitoring.

tmasternak:
@mikeminutillo can you have a look. I think we addressed all of your comments.

src/ServiceControl.Monitoring/Infrastructure/BreakdownRegistry.cs
```diff
@@ -0,0 +1,41 @@
+namespace ServiceControl.Monitoring.Infrastructure
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+
+    public abstract class BreakdownRegistry<BreakdownT>
```
mikeminutillo: This is optimized for reading at the expense of locking on every write. Is that assumption OK?

Scooletz: That's correct. Now I'm thinking, whether it should be something else. We call Record for every message. On the other hand, it's unlikely to breach 10k reporting messages per sec, right? (the number=average lock throughput)

Scooletz: I made the write side much lighter.

PR #55, opened 2017-09-05
Minor fixes after integration tests

tmasternak:
Those are minor fixes:
 * Handling situation when message-type registry does not hold mapping for given endpoint-name
 * Tracking endpoint instances and endpoint message types based on `TaggedLongValueOccurrence` message
 * Removing unused `Occurrence` and `LongValueOccurence` messages


src/ServiceControl.Monitoring/Infrastructure/BreakdownRegistry.cs
```diff
@@ -37,7 +37,7 @@ public void Record(BreakdownT breakdown)
 
         public IEnumerable<BreakdownT> GetForEndpointName(string endpointName)
         {
-            return lookup[endpointName];
+            return lookup.ContainsKey(endpointName) ? lookup[endpointName] : new List<BreakdownT>();
```
Scooletz: This is a concurrently accessed field. Also, the lookup can be replaced by `TryGet`. To address these two:

```
if (lookup.TryGet(endpointName, out breakdowns))
...
```

Scooletz:
Also, the build is broken with one unneeded `using` statement.

PR #57, opened 2017-09-06
Exposing metrics digest via endpoint details API

tmasternak:
## Overview
This PR exposes metrics data digest (last value and average for selected period) under `endpoint-details` url.

## Example
url: `http://localhost:33633/monitored-endpoints/Subscriber?history=5`

data:

``` json
{
    "digest": {
        "metrics": {
            "processingTime": {
                "latest": 0,
                "average": 1289.1179138321995
            },
            "criticalTime": {
                "latest": 0,
                "average": 17266.559718969555
            },
            "retries": {
                "latest": 0,
                "average": 1.35
            },
            "queueLength": {
                "latest": 1487,
                "average": 1390
            },
            "throughput": {
                "latest": 0,
                "average": 0.49
            }
        }
    },
    "instances": [ ],
    "messageTypes": []
}
```


src/ServiceControl.Monitoring/Http/Diagrams/MonitoredEndpointDigest.cs
```diff
@@ -0,0 +1,9 @@
+namespace ServiceControl.Monitoring.Http.Diagrams
+{
+    using System.Collections.Generic;
+
+    public class MonitoredEndpointDigest
+    {
+        public Dictionary<string, MonitoredEndpointMetricDigest> Metrics { get; set; } = new Dictionary<string, MonitoredEndpointMetricDigest>();
```
Scooletz: This allows extensibility on the server side, which is great. I'm thinking now, should we write a test for public http API that would ensure that right metrics are in there? Just a thought.

tmasternak: @Scooletz I like the idea. I think we could test other dictionaries (inside instances, metric-types and endpoint) as well. Separate PR?

PR #61, opened 2017-09-21
[WIP] Receiving Queue Length reports with tagget values instead of JSON report

Scooletz:
This PR introduces an infrastructure that supports receiving Queue Length not only with JSON Metrics.NET like format but also tagged values.


src/ServiceControl.Monitoring/QueueLength/TaggedLongValueOccurrenceHandler.cs
```diff
@@ -0,0 +1,53 @@
...
+            {
+                case QueueLengthSent:
+                    queueLengthStore.UpdateSentSequence(message.TagValue, value);
+                    break;
+                case QueueLengthReceived:
+
+                    var virtualQueueId = new VirtualQueueId
+                    {
+                        EndpointName = instanceId.EndpointName,
+                        QueueName = WHAT,
```
Scooletz: This is a breadcrumb I left. Could somebody help with figuring out how this should be set up? Also, answering this is required to make the PR reviewed properly reviewed (it's worth it only when we can obtain the queue name somehow). Otherwise it could require us to send two tags or sth similar https://github.com/Particular/ServiceControl.Monitoring.Data/pull/7

Scooletz: @tmasternak Could you take a look at it?

Scooletz:
This closed for now as @Particular/metrics-maintainers chose another path address the reporting.

PR #62, opened 2017-09-27
Default input queue changed to `Particular.Monitoring`

tmasternak:
This PR changes default input queue. The new value is the same `SCMU` will assign to newly created  monitoring instance.
With this change it's easier to setup development environment. One can have single configuration for monitored endpoint and use either `SCMU` installed or development version of monitoring instance.


Scooletz:
Doesn't it require changes in doco? How do pass the name of the queue to the Metrics' plugin

tmasternak:
AFAIK the part of the API that sends data to monitoring instance is not yet covered in the docs. I've did some searching now and did not find anything.

Scooletz:
Alrighty then!

PR #63, opened 2017-09-28
Current interval data not included in interval returned by IntervalStore

tmasternak:
This PR changes the interval that are returned by `IntervalStore`. It will not include the current interval. This makes the latest returned value a bit more stable.


src/ServiceControl.Monitoring/Infrastructure/IntervalsStore.cs
```diff
@@ -57,7 +57,7 @@ public Measurement(TimeSpan intervalSize, int numberOfIntervals)
             // ReSharper disable once SuggestBaseTypeForParameter
             public void ReportTimeIntervals(DateTime now, IntervalsBreakdown item)
             {
-                var epoch = GetEpoch(now.Ticks);
+                var epoch = GetEpoch(now.Ticks) - 1;
```
Scooletz: Magic const. Can we change it to

```
var currentEpoch = GetEpoch(now.Ticks);
var previousEpoch = currentEpoch - 1;
```

or sth similar to make it more explicit?

tmasternak: I've keep the `epoch` instead of `previousEpoch` as it's used inside a look and there it means epoch to be included in reported data.

Scooletz:
I wanted to remind us that because of the "update" on this branch it contains merge commit **from develop** the commit history is a bit complex. It'd be great if maintainers could "squash and merge" it.

tmasternak:
> I wanted to remind us that because of the "update" on this branch it contains merge commit from develop the commit history is a bit complex

So `Squash and merge` will keep history clean right?

Scooletz:
Yes @tmasternak

PR #65, opened 2017-10-05
Retries and Throughput avg counted using unique aggregation intervals

tmasternak:
## Overview
This PR fixes the way we calculate the averages for Retries and Throughput metrics. Previously the number of intervals used in average formulas was taken as a total count of intervals. That caused problems when aggregating data came from overlapping intervals. In such case the number of intervals calculated was bigger when number of unique intervals and caused avg to be smaller than expected.

## Sample
Let's say we have following data about Processing Time (Note: Throughput is calculated based on that):
  * Endpoint A - Instance 1: [ [t:0, total: 10], [t:1, total: 10]]  // t is start time for interval
  * Endpoint A - Instance 2: [ [t:1, total: 5], [t:2, total: 5]] 

Currently average for Throughput for **logical Endpoint A** will be calculated as:

```
 avg = (10+10+5+5)/4  // 4 because this is total number of intervals
```

the correct calculation should be

```
avg = (10+10+5+5)/3 // 3 because 2nd interval for Instance 1 and 1st for Instance 2 is in fact the same one
```

Fixes https://github.com/Particular/LaunchWave.DevOpsAdvancedMonitoring/issues/279


tmasternak:
Please use `Squash and merge` when merging this.

Scooletz:
Could you provide a bigger example for "data came from overlapping intervals", how is it possible? Where's the query returning the overlapping intervals?

Scooletz:
@tmasternak ☝️

tmasternak:
@Scooletz description updated with Sample section

Scooletz:
Thank you @tmasternak Now this provides a clear and full picture of this fix.

PR #66, opened 2017-10-06
Turning off any http caching for proxies and clients

tmasternak:
## Overview
This PR adds HTTP response headers to prevent proxies and clients from caching responses.

Partial fix for: https://github.com/Particular/LaunchWave.DevOpsAdvancedMonitoring/issues/270


Scooletz:
These parameterless API calls and caching ;-)

src/ServiceControl.Monitoring/Http/ApiModule.cs
```diff
@@ -9,7 +10,11 @@ protected ApiModule()
                 .WithHeader("Access-Control-Allow-Origin", "*")
                 .WithHeader("Access-Control-Allow-Methods", "POST,GET")
                 .WithHeader("Accept", "application/json")
-                .WithHeader("Access-Control-Allow-Headers", "Accept, Origin, Content-type"));
+                .WithHeader("Access-Control-Allow-Headers", "Accept, Origin, Content-type")
+                .WithHeader("Expires", "Tue, 03 Jul 2001 06:00:00 GMT")
+                .WithHeader("Last-Modified", DateTime.Now.ToUniversalTime().ToString("R"))
+                .WithHeader("Cache-Control", "max-age=0, no-cache, must-revalidate, proxy-revalidate")
```
WilliamBZA: Need to set `no-store` as well as a `no-chache` pragma header

``` 
.WithHeader("Cache-Control", "max-age=0, no-cache, must-revalidate, proxy-revalidate, no-store")
.WithHeader("Pragma", "no-cache")
```

Worth noting that no client is required to respect these values either. We should be cachebreaking based on e-tags as well, but that's not worth the effort at this stage.

tmasternak: * `Pragma: no-cache` is not a response but request header https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Pragma
 * `Cache-Control: no-store` this says caching proxies are not allowed to store response content to disk. I always assumed that this does not make sense with `no-cache` but I might be wrong here.

WilliamBZA: Cache management is a giant troll: https://vimeo.com/68331691

WilliamBZA: The correct fix is proper cache breaking strategy. It's on the SP list of things to do :(

tmasternak:
I'm not an expert in Cache-Breaking as well. I've added the headers mentioned by @WilliamBZA to be on the safe side.

PR #69, opened 2017-10-23
1 min history period

tmasternak:
## Overview
This PR introduces changes that enables more frequent data refreshing on the SP end. It does it by introduction following changes:
  * 1 min history period divided into 30 intervals (interval size is 2 sec),
  * Providing variable reporting time delay for historical periods. Right now for 1 min interval data will be reported with 10 sec delay for others it will be as before i.e. data for current interval will not be reported.

Those parameters are somehow arbitrary and have been driven by test performed at the UI level. The interval size should not be made smaller though as current queue length data is snapshoted every second.

## Memory consumption tests
During implementation a simple verification of memory consumption have been made when switching from **20** to **40** intervals per history period.

### Scenario

**NOTE: those measurements where taken when intervals number has been changed to 40 for all periods. This PR makes this change only for single period - 1 min one.**

The scenario of the test was:
   * Run performance test `GetMonitoredEndpointsQueryTest` with biggest configuration (100 endpoints and 10 instances per endpoint)
   * Snapshot heap  stats in VS
   * Calculate total size of objects connected to SC.Monitoring and **not the test infrastructure** 
### Results
The memory consumption has grown from **25 MB** to **44 MB**

   * For **20** intervals per history period the total size was ~**25** MB:
image
    * For **40** intervals per history period the total size was ~**44** MB
image


src/ServiceControl.Monitoring/Infrastructure/HistoryPeriod.cs
```diff
@@ -21,7 +22,7 @@ static HistoryPeriod()
         HistoryPeriod(TimeSpan value)
         {
             Value = value;
-            NumberOfIntervals = 20; //Number of points on the graph
+            NumberOfIntervals = 40; //Number of points on the graph
```
Scooletz: This doubles the memory needed for storing metrics, right? Any measurements were taken?

Scooletz: @tmasternak ☝️

tmasternak: You mean tests that would validate that memory consumption changes as expected?

Scooletz: I mean any test, even doing a smoke test. This looks like a not so small change in terms of memory consumption. Or maybe, it's nothing that big?

tmasternak: @Scooletz I've updated issue description with memory consumption tests. IMO the memory consumption growth of 19 MB is something we can live with.

I'm more concerned with increased times for data queries. The mean went over 50 ms and is around 89 ms on the build machine.

tmasternak: @Scooletz after thinking a bit more on the number of intervals I've realized that we need more only for the smallest interval (1 min). 

I've made the intervals per history period 20 for any period larger than 1 min. Hopefully that won't break the perf tests we have. 

I propose to keep the memory measurements in the description for future reference.

Scooletz: Interesting finding. Still 89ms for a service accessed by a single user without too long network hops seems ok-ish for a reporting tool. 

Nice findings!

src/ServiceControl.Monitoring.AcceptanceTests/Tests/ApiIntegrationTest.cs
```diff
@@ -42,7 +42,7 @@ string GetString(string url)
             return httpClient.GetStringAsync(url).GetAwaiter().GetResult();
         }
 
-        string MonitoredEndpointsUrl = "http://localhost:1234/monitored-endpoints";
+        string MonitoredEndpointsUrl = "http://localhost:1234/monitored-endpoints?history=1";
```
mikeminutillo: Why is this needed? Is this no longer the default? I'm not seeing this change in the PR.

src/ServiceControl.Monitoring/Infrastructure/HistoryPeriod.cs
```diff
@@ -8,36 +8,35 @@ public class HistoryPeriod
...
         {
-            All = new List<HistoryPeriod>()
+            All = new List<HistoryPeriod>
             {
-                new HistoryPeriod(TimeSpan.FromMinutes(5)),
-                new HistoryPeriod(TimeSpan.FromMinutes(10)),
-                new HistoryPeriod(TimeSpan.FromMinutes(15)),
-                new HistoryPeriod(TimeSpan.FromMinutes(30)),
-                new HistoryPeriod(TimeSpan.FromMinutes(60))
+                new HistoryPeriod(TimeSpan.FromMinutes(1), 40),
```
mikeminutillo: This results in 1.5 second buckets. Was 40 buckets just an arbitrary doubling of the previous number? Would it make sense to have 30 2-second buckets or even 60 1-second buckets? Do does that mess with the UI?

src/ServiceControl.Monitoring.AcceptanceTests/Tests/ApiIntegrationTest.cs
tmasternak: Side note: I think that this PR is not focused enough :/. I've added changes to tests to make them run faster and be able to run only subset easily. 

The reason why I've added the `?history=1` is that the  current aggregation interval is not reported so for integration tests it means we have to wait until current interval has passed to get any data when doing queries. The shorter history period the shorter interval -> I've switched to history period of 1 min.

src/ServiceControl.Monitoring/Infrastructure/HistoryPeriod.cs
tmasternak: This is a good point. What we want to do is get on a call and tweak UI and SC.Montoring at the same time to get best results. 

Until than I keep this as `WIP`

PR #70, opened 2017-10-23
Scheduled retries aggregated to rate per second

tmasternak:
## Overview
This PR changes how Scheduled Retries are aggregated. It uses the same approach as Throughput i.e. calculates rate or measurements per second.

Tests for retry aggregations have been removed as aggregation is done with `AggregateTotalMeasurementsPerSecond` which is already covered.


src/ServiceControl.Monitoring/Http/Diagrams/DiagramApiModule.cs
```diff
@@ -28,7 +28,7 @@ public MonitoredEndpointsModule(IProvideBreakdownBy<EndpointInstanceId>[] metric
             {
                 CreateMetric<EndpointInstanceId, ProcessingTimeStore>("ProcessingTime", IntervalsAggregator.AggregateTimings),
                 CreateMetric<EndpointInstanceId, CriticalTimeStore>("CriticalTime", IntervalsAggregator.AggregateTimings),
-                CreateMetric<EndpointInstanceId, RetriesStore>("Retries", IntervalsAggregator.AggregateRetries),
```
mikeminutillo: Can `IntervalsAggregator.AggregateRetries` also be 🔥 ?

tmasternak: done.

tmasternak:
@Scooletz note on test removal added in the description.

tmasternak:
@mikeminutillo review changes are in place

PR #72, opened 2017-10-27
All history periods with 60 intervals

tmasternak:
## Overview
This PR introduces increased number of intervals to all history periods. Previously it was 20 now it's 60.

Memory footprint can be estimated based on measurements made by #69.

The change has been made to enable better responsiveness in SP. Currently when switching from 1 min history period to 5 min there is too big of a difference in refresh rates (1 sec vs 15 sec). 

**IMPORTANT: this PR increases allowed report query time for the biggest case from 50 ms to 120 ms.**


tmasternak:
@Scooletz thoughts on the performance degradation?

tmasternak:
@Scooletz I agree.

Scooletz:
Ping @Particular/servicecontrol-maintainers ? ;-)

WilliamBZA:
I dont think the ServiceControl maintainers have the required context to make that decision.

WilliamBZA:
Do we know where the extra time is spent? Serialization or reading the data?

Scooletz:
@WilliamBZA Reading data with LINQ. The query part was not optimized and begs for polishing, which still can be postponed as it acts reasonably well for a local SPA app.

WilliamBZA:
Sounds reasonable as something that can be optimized further if (once?) it proves itself to be a real problem. No need for concern now though.

Scooletz:
Does it mean that as a TF member I can hit this beautiful "merge" button? ;-) @WilliamBZA

WilliamBZA:
asd

PR #75, opened 2017-11-17
150 ms query limit for biggest performance test

tmasternak:
## Overview
Recently 2 builds of develop failed with largest performance tests query limit being 121 and 125 ms. This caused the builds to fail as the limit was set to 120 ms.

There were was only a single change made in the meantime (since the tests where consistently passing) - in the booting sequence that should not affect the performance.

It looks that either the previous value was right on the border or the performance characteristics of the build agent changed.


tmasternak:
@Scooletz issue created #76

Issue #76, opened 2017-11-17
Optimize query side


tmasternak:
## Overview
The biggest performance tests are passing with current query limit set to 150 ms. We already increased this bar two times and need to look on bringing this number back down.

tmasternak:
This has not been seen to cause any troubles to the users.

PR #78, opened 2017-11-21
Structured message type in API

tmasternak:
## Overview
This PR changes values of message types returned by the API. Instead of `AssemblyQualifiedName` it returns a structured json object with properties per each part of `AssemblyQualifiedName`.

## Assumptions
Parsing is based on the assumption that the value that is parsed is either:

   * Empty string - this happens when we don't get message type with metrics data e.g. control messages
   * Valid `EnclosedMessageType` [header value](https://docs.particular.net/nservicebus/messaging/headers#serialization-headers-nservicebus-enclosedmessagetypes). Which means that it's either:
     * Type.Name or
     * Type.AssemblyQualifedName 

**If the value does not fall into one of the categories above the value as a whole is considered a type name**.

This is part of https://github.com/Particular/LaunchWave.DevOpsAdvancedMonitoring/issues/242


src/ServiceControl.Monitoring/Http/Diagrams/DiagramApiModule.cs
```diff
@@ -196,10 +196,7 @@ static MonitoredEndpoint[] GetMonitoredEndpoints(EndpointRegistry endpointRegist
         static MonitoredEndpointMessageType[] GetMonitoredMessageTypes(MessageTypeRegistry registry, string endpointName)
         {
             return registry.GetForEndpointName(endpointName)
-                .Select(mt => new MonitoredEndpointMessageType
-                {
-                    MessageType = mt.MessageType
-                })
+                .Select(mt => MonitoredEndpointMessageType.FromTypeName(mt.MessageType))
```
Scooletz: Could this be moved to `MessageTypeRegistry`? It would run transformation only once for each message type. No need to do the computation on every access.

src/ServiceControl.Monitoring/Http/Diagrams/MonitoredEndpointMessageType.cs
```diff
@@ -1,11 +1,67 @@
...
+            //      What it means that when non empty the value is either TypeName or AssemblyQualifiedName
+
+            if (string.IsNullOrEmpty(typeName))
+            {
+                return new MonitoredEndpointMessageType();
+            }
+
+            var typeParts = typeName.Split(new []{','}, StringSplitOptions.RemoveEmptyEntries);
+
+            if (typeParts.Length > 1)
```
Scooletz: I'm scared that we omit some kind of condition in here. Have you considered using `AssemblyName` instead of a custom parser in here?

src/ServiceControl.Monitoring.Tests/API/MonitoredEndpointMessageTypeTests.cs
```diff
@@ -0,0 +1,77 @@
...
+    using NUnit.Framework;
+
+    public class MonitoredEndpointMessageTypeTests
+    {
+        [Test]
+        public void Parsing_empty_string_returns_empty_type_name()
+        {
+            var result = MonitoredEndpointMessageType.FromTypeName("");
+
+            Assert.IsNull(result.TypeName);
```
Scooletz: Shouldn't it assert the rest of properties as well?

src/ServiceControl.Monitoring/Http/Diagrams/MonitoredEndpointMessageType.cs
tmasternak: That's a good idea. I haven't consider using `AssemblyName` for the part after first comma. Let me try that.

src/ServiceControl.Monitoring/Http/Diagrams/DiagramApiModule.cs
tmasternak: I'm not sure `MessageTypeRegistry` is a proper place. `MonitoredEndpointMessageType` is part of api contract and we should not add dependency to it from infrastructural parts. 

I do think that we should cache the values though. I'll create a dedicated caching parser for this purpose.

tmasternak:
@Scooletz I've addressed the review comments.

src/ServiceControl.Monitoring/Http/Diagrams/DiagramApiModule.cs
```diff
@@ -17,7 +17,7 @@ public class MonitoredEndpointsModule : ApiModule
         /// Initializes the metric API module.
         /// </summary>
         // ReSharper disable SuggestBaseTypeForParameter
-        public MonitoredEndpointsModule(IProvideBreakdownBy<EndpointInstanceId>[] metricByInstance, IProvideBreakdownBy<EndpointMessageType>[] metricByMessageType, EndpointRegistry endpointRegistry, EndpointInstanceActivityTracker activityTracker, MessageTypeRegistry messageTypeRegistry)
+        public MonitoredEndpointsModule(IProvideBreakdownBy<EndpointInstanceId>[] metricByInstance, IProvideBreakdownBy<EndpointMessageType>[] metricByMessageType, EndpointRegistry endpointRegistry, EndpointInstanceActivityTracker activityTracker, MessageTypeRegistry messageTypeRegistry, MonitoredEndpointMessageTypeParser messageTypeParser)
```
Scooletz: I see the injection point but fail to find registration point. Does it meant that `MonitoredEndpointMessageTypeParser` will be a singleton. If not, the caching that is instance based won't work.

src/ServiceControl.Monitoring/Http/Diagrams/MonitoredEndpointMessageTypeParser.cs
```diff
@@ -0,0 +1,68 @@
...
+    using System.Reflection;
+    using NServiceBus.Logging;
+
+    public class MonitoredEndpointMessageTypeParser
+    {
+        ConcurrentDictionary<string, MonitoredEndpointMessageType> messageTypes = new ConcurrentDictionary<string, MonitoredEndpointMessageType>();
+
+        public MonitoredEndpointMessageType Parse(string typeName)
+        {
+            if (!messageTypes.ContainsKey(typeName))
```
Scooletz: `TryGetValue` instead of `ContainsKey` && additional lookup at the end.

src/ServiceControl.Monitoring/Http/Diagrams/DiagramApiModule.cs
tmasternak: The registration is done by the containers. All types are [singletons](https://github.com/Particular/ServiceControl.Monitoring/blob/develop/src/ServiceControl.Monitoring/EndpointFactory.cs#L150)

src/ServiceControl.Monitoring/Http/Diagrams/MonitoredEndpointMessageTypeParser.cs
```diff
@@ -12,13 +12,15 @@ public class MonitoredEndpointMessageTypeParser
...
 
-                messageTypes.TryAdd(typeName, messageType);
+            if (messageTypes.TryGetValue(typeName, out messageType))
+            {
+                return messageType;
             }
 
+            messageTypes.TryAdd(typeName, FromTypeName(typeName));
+
             return messageTypes[typeName];
```
Scooletz: If that's a cached item, you could return also the one that you `TryAdd` and remove accessing dictionary altogether. This is not heavy requirement as the hot path (reading) issue has been addressed.

src/ServiceControl.Monitoring/Http/Diagrams/DiagramApiModule.cs
Scooletz: Lovely!

tmasternak:
@Scooletz ok this gets interesting :P. The previous build failed due to the fact that the `MonitoredEndpointInstanceMessageType` has a single instance per message type received by monitoring instance. 
What it means is that the same instance is shared between different client queries. Now that would work fine if those instances where immutable but they are not :(. They hold `Metrics` dictionaries. As a result what is possible is two clients quering for metrics of the same message type and racing over access to `Metrics` property (this is what happened for the tests that failed).

What I propose to do is back-off from caching and tackle it as part of #76. The arguments for this are:
 * We don't know if caching is the query bottleneck and thing to tackle first,
 * If we cache message types we probably want to cache endpoint instances,
 * We need to slightly redesign api responses to curve out immutable parts.

Thoughts?

Scooletz:
@tmasternak Agreed. Sorry I missed this.

tmasternak:
@Scooletz perf tests for query side fail with 200 ms. I don't want to make it higher one more time. 

I propose to tweak this PR until we meet perf criteria. If it requires caching I say we introduce caching.

Scooletz:
I think I found the issue: https://github.com/Particular/ServiceControl.Monitoring/blob/develop/src/ServiceControl.Monitoring/Infrastructure/IntervalsStore.cs#L111-L148

we can do all the calculations before entering Writer lock, can't we? Then, create an array with aggregated items, then enter write lock, loop over, end it.

Scooletz:
Please consider https://github.com/Particular/ServiceControl.Monitoring/pull/79 before applying fixes to this @tmasternak

PR #79, opened 2017-11-22
[WIP] Optimizations

Scooletz:
This PR lower the overhead of the coarse grained ReaderWriterLock. It's done by aggregating data first in `workingCopyOfIntervals` and then locking the common structure with `RWL` for writing aggregated data. One `GetEpoch` and one `.Ticks` are inlined as well.


Scooletz:
This requires some further discussion and analysis. I'm not sure if the target for 20ms is not to low.

tmasternak:
> I'm not sure if the target for 20ms is not to low.

Should that be 200 ms?

tmasternak:
I like the approach. Here are some brainstorming ideas:
  * What if we kept two copies of `Intervals` one for writers and second for readers. In addition we could have an async task that would copy first into the second (every `n` ms)? So effectively we would trade data freshness for congestion on write side,
  * Do we need that much consistency on the read side. What if we during queries we would copy `Entry` struct one at a time in consistent way? This problematic because we still need consistent view of `Epoch`, `TotalMeasurements` and `TotalValue`. I was thinking about packing that into int64 and using `Interlocked` but I'm not sure that is possible,
  * Let's have a single thread managing the intervals connected to writers and readers via bounded in-memory queue. We can make queue synchronization less costly than RWL.

Those are only possible ideas. Let's have a discussion :).

Scooletz:
If we want to change it, I'd go with the second. Packing is not possible. To `CompareExchange` we'd need to make it `class`. Unfortunately for us, there's no `Interlocked128` which is available on majority of CPUs today. Let's sync on Monday.

Scooletz:
This PR is closed as performance improvements mentioned in here increase memory consumption without providing the actual solution to the failing test. The failing test is a result of a non-delayed reads, flooding stores with a constant stream of queries not separated in time.

Scooletz:
Ping @tmasternak

PR #83, opened 2017-11-27
More realistic performance tests

Scooletz:
This PR fixes the way we were querying stores in the performance tests. 

Queries were not properly delayed, issuing a query one after another, without any delay as it would happen with an SP instance running and pulling for data. Now, the test parameter that has specified the value of the delay is taken into consideration.


Scooletz:
@tmasternak So it looks that making this test more realistic did not change the behavior, and still, we're able to breach the query time. What do you propose?

tmasternak:
@Scooletz I've still slightly over 20 ms on the reporters side. I propose to increase it to 25 ms (commit made).

Scooletz:
@tmasternak looks 👍 I cannot review my own change :P

PR #85, opened 2017-11-30
Instance metrics are aggregated by EndpointName and InstanceId

tmasternak:
## Overview
This PR fixes a bug in the instance metric aggregation logic. In the previous version if there were two instances in two different logical endpoints but with the same `instanceId` the aggregation would combine interval data from both of those.

This fix makes sure that before aggregation we gather instance intervals only in scope of a single logical endpoint.


tmasternak:
@Scooletz test added.

PR #98, opened 2018-01-09
Support Amazon SQS transport

Scooletz:
This PR introduces the support for Amazon SQS for the `ServiceControl.Monitoring` by providing:
- [x] smoke tests for SQS transport
- [x] SQS transport override for `3.1` version of it, enabling to set the region via env variable or the connection string
- [x] packaging


danielmarbach:
Cool. We wondered if it is some kind of convention. We will fix it later

Am 11.01.2018 13:00 schrieb "Tomasz Masternak" <notifications@github.com>:

Scooletz:
Will fix

Scooletz:
@tmasternak I amended the mapping.

Scooletz:
@tmasternak So it looks that the transport override fails to get loaded.

I wrote a local test with `Type.GetType(string)` but it worked. Is there any other magic that is required to get the transport loaded with its assembly qualified name 

```
"ServiceControl.Transports.AmazonSQS.ServiceControlSqsTransport, ServiceControl.Transports.AmazonSQS"
```

Scooletz:
Marking as WIP till we clarify the loading error.

danielmarbach:
@tmasternak can you please review. It works now on my machine (TM)

Scooletz:
Please `Squash and Merge` after review. It had its branch updated from `develop`.

Scooletz:
Ping @Particular/servicecontrol-maintainers This is green and ready for a review.

src/ServiceControl.Transports.AmazonSQS/ServiceControlSqsTransport.cs
```diff
@@ -0,0 +1,47 @@
...
+            // SQS doesn't support connection strings so pass in null.
+            return base.Initialize(settings, null);
+        }
+
+        static string PromoteEnvironmentVariableFromConnectionString(DbConnectionStringBuilder builder, string connectionStringKey, string environmentVariableName)
+        {
+            if (builder.TryGetValue(connectionStringKey, out var value))
+            {
+                var valueAsString = (string) value;
+                Environment.SetEnvironmentVariable(environmentVariableName, valueAsString, EnvironmentVariableTarget.Process);
```
mikeminutillo: Are there special permissions required to do this? If we run the SC windows service as a network user or local machine will it be able to update the process environment variables? I think it should be OK, just checking.

danielmarbach: Not thatI know. Will check but the assumption was a process has always right to write env stuff for itself

danielmarbach: @mikeminutillo couldn't find anything mentioning the process cannot set its own variables. The good thing about this approach is that the env variables are bound to the lifetime of the process, if the process goes the env variables are gone as well.

PR #99, opened 2018-01-11
Forwarding topology type will be deprecated, use settings instead

danielmarbach:
The types used here will be internalized in the future version of the transport. That's why there is already an obsoletion message that we had to suppress. This is the more future proof way


danielmarbach:
@Scooletz as discussed

tmasternak:
@danielmarbach can you give some more context? Why do we want to have this instead of previous version?

PR #101, opened 2018-01-15
Release 1.1.0

Scooletz:


src/ServiceControl.Monitoring.SmokeTests.SQS/EndpointTemplates/TestSuiteConstraints.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        bool SupportsOutbox { get; }
+
+        IConfigureEndpointTestExecution TransportConfiguration { get; }
+
+        IConfigureEndpointTestExecution PersistenceConfiguration { get; }
+    }
+
+    // ReSharper disable once PartialTypeWithSinglePart
+    public class TestSuiteConstraints : ITestSuiteConstraints
```
danielmarbach: Should we align this with https://github.com/Particular/NServiceBus.AmazonSQS/blob/develop/src/AcceptanceTests/TestSuiteConstraints.cs ? I think we should

PR #104, opened 2018-01-18
Faster performance tests

Scooletz:
This PR removes some of the mostly redundant test cases, as they are covered by the remaining suite. The picture below shows top ~misbehaving~  slow tests:

top

The change introduced by this PR:
- before, tests take: ~**13 minutes** to complete
- after, tests take: ~**6 minutes** to complete


Scooletz:
@mikeminutillo I left the suite with a representative for every test. I don't claim I chose the best ;-)

PR #107, opened 2018-01-19
Allow skipping queue creation

mikeminutillo:
Part of #106

### DIFF
--- a/src/ServiceControl.Monitoring/Hosting/Commands/SetupCommand.cs
+++ b/src/ServiceControl.Monitoring/Hosting/Commands/SetupCommand.cs
@@ -9,6 +9,10 @@ namespace ServiceControl.Monitoring
         {
             var endpointConfig = EndpointFactory.PrepareConfiguration(settings);
             endpointConfig.EnableInstallers(settings.Username);
+            if (settings.SkipQueueCreation)
+            {
+                endpointConfig.DoNotCreateQueues();
+            }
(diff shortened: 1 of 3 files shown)


src/ServiceControl.Monitoring/Hosting/Commands/SetupCommand.cs
```diff
@@ -9,6 +9,10 @@ public override Task Execute(Settings settings)
         {
             var endpointConfig = EndpointFactory.PrepareConfiguration(settings);
             endpointConfig.EnableInstallers(settings.Username);
+            if (settings.SkipQueueCreation)
```
WilliamBZA: It's a pity we can't invert the logic easily. I.e. have the default be to not create the queues.

mikeminutillo: @WilliamBZA Why is that an advantage? At the moment we only create queues if you run with the `--setup` flag. It's assumed that part of the setup of the endpoint is to migrate any internal data stores, create folders on disk, create queues, etc.

WilliamBZA: Opt in feels better to me than opt out - especially for things that have a material impact on the environment. No string feelings about it though.

mikeminutillo:
Rebased on develop

PR #110, opened 2018-02-01
EndpointInstanceId equality not dependent on InstanceName

tmasternak:
## Overview
This PR changes the equality relation between `EndpointInstanceId` instances. Previously it took into consideration `InstanceName` which caused faulty behavior.

In scenarios when users specified custom display names via core api:

```
  endpointConfiguration.UniquelyIdentifyRunningInstance()
                .UsingCustomIdentifier(new Guid("1C62248E-2681-45A4-B44D-5CF93584BAD6"))
                .UsingCustomDisplayName("original-instance-rename");
```

and changed `DisplayName` during instance lifetime (keeping `CustomIndentifier` same) they would end up with two entries representing same endpoint instance (for the old name and the new one).

/cc: @Scooletz thx for help!

### DIFF
--- a/src/ServiceControl.Monitoring/Infrastructure/BreakdownRegistry.cs
+++ b/src/ServiceControl.Monitoring/Infrastructure/BreakdownRegistry.cs
@@ -6,7 +6,7 @@ namespace ServiceControl.Monitoring.Infrastructure
 
     public abstract class BreakdownRegistry<BreakdownT>
     {
-        HashSet<BreakdownT> breakdowns = new HashSet<BreakdownT>();
+        Dictionary<BreakdownT, BreakdownT> breakdowns = new Dictionary<BreakdownT, BreakdownT>();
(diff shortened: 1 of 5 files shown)


src/ServiceControl.Monitoring/Infrastructure/BreakdownRegistry.cs
```diff
@@ -21,25 +21,35 @@ public void Record(BreakdownT breakdown)
...
                 {
-                    lookup = breakdowns.ToArray()
+                    lookup = breakdowns.Values.ToArray()
                         .GroupBy(b => endpointNameExtractor(b))
                         .ToDictionary(g => g.Key, g => (IEnumerable<BreakdownT>)g.Select(i => i).ToArray());
                 }
             }
         }
 
+        protected virtual bool AddBreakdown(BreakdownT breakdown, Dictionary<BreakdownT, BreakdownT> existingBreakdowns)
```
Scooletz: What a beautiful usage of a `virtual` method!

PR #113, opened 2018-04-09
[WIP] Add instance key

mikeminutillo:
Fixes https://github.com/Particular/NServiceBus.Metrics.ServiceControl/issues/27

Splits the message source key into Logical Source and Instance Key (based on https://github.com/Particular/NServiceBus.Metrics.ServiceControl/pull/28) and aggregates the state of Instances under Logical Message Source (both for sending and receiving).

If a separate instance key is not provided, then treat it as empty. i.e. All sessions without a unique key are considered the same session (which is what it was doing before).

This should allow each instance to restart (resetting it's send counter) and `immediately contribute to queue length even if no messages from the new session have been received.


mikeminutillo:
@Scooletz hold off for now

mikeminutillo:
Left on hold for now. This fixes an issue with the queue length estimate. It won't be needed once we have a native queue length implementation for each transport.

tmasternak:
We are going to switch to native queue length. Estimated algorithm will be removed.

PR #114, opened 2018-04-11
[SPIKE] Native queue length

mikeminutillo:
What if the endpoints could send us the length of all of their queues (tagged by physical queue address)?

Where physical input queues overlap (competing consumer style) we take the max reported value for the interval. Where they don't (Sender-Side-Distribution) we take the sum of the reported values for the interval.

Note that we could go even further and break this down by instance, and we could then further break down the instance and show each of it's individual input queues.

OR we could show a breakdown at the logical level level showing all of the input queues that the logical endpoint uses.


Scooletz:
@mikeminutillo This approach would require every endpoint to get to all the queues lenghts. This means that we'd need to extend the transport on the client side, releasing  a bunch of packages with changes made to address it. Additionally, for the scaled out endpoints (competing consumers) you'd obtain the length value on each instance. I'm not saying it's a heavy-compute operation, just noticing that we'd need a lot of queries against the queue.

☝️ If any of this is incorrect, please raise it :)

I'm looking forward to the outcome of the [native queue length](https://github.com/Particular/LaunchWave.DevOpsAdvancedMonitoring/issues/539) as this looks most promising to me.

tmasternak:
Spike is over.

PR #116, opened 2018-04-12
[SPIKE] Centralized native queue length

tmasternak:
## Overview
This spike shows how to implement native queue length metric that is queried by a monitoring instance. It supports SQL transport only.

Main features:
  * The physical queue names for logical endpoint are pulled from metric reports sent by each endpoint
  * Each transport (currently only sql) provides implementation of `INativeQueueLengthProvider`
  * `QueueLength` passes all discovered queue names to the native provider and asks for latest know values it knows,
  * The queue length metrics store will use native value for queue length if one is present otherwise it will fallback to value provided by estimation algorithm.

## Shortcommings
 * Currently the queue length provider is trivial and does not do any batching of requests,
 * This does not provide mulit broker/instance setup but could be implemented e.g. by extending monitoring plugin api with something like `.EnableMetrics(c => c.TransportBrokerId("broker-id"))`


src/ServiceControl.Transports.SQLServer/NativeQueueLengthProvider.cs
```diff
@@ -0,0 +1,94 @@
...
+                .ToLookup(kv => kv.Value, kv => kv.Key)
+                .ToDictionary(g => g.Key, g => g.ToArray().Sum(qn =>
+                                                            {
+                                                                queueSizes.TryGetValue(qn, out var length);
+                                                                return length;
+                                                            })
+                );
+
+            return result;
+        }
```
mikeminutillo: I think we'll still need to break this down on EndpointInstanceId _and_ physical queue name. If we don't then we'll double count shared queues in competing consumer mode.

tmasternak:
Spike is over.

PR #117, opened 2018-04-16
[SPIKE] Centralized Native Queue Length for Learning Transport

mikeminutillo:
Showing a second implementation of https://github.com/Particular/ServiceControl.Monitoring/pull/116

NOTE: The non-native estimate was not working properly for this as it overwrites values for the endpoint name without the instance id. 

Either we need to not pass in the instance id to the TrackQueue call or we'll need a better breakdown algorithm. Otherwise we will be duplicating instance data for every competing consumer


src/ServiceControl.LearningTransport/LearningTransportNativeQueueLengthProvider.cs
```diff
@@ -0,0 +1,88 @@
...
+
+    public class LearningTransportNativeQueueLengthProvider : INativeQueueLengthProvider
+    {
+        string rootFolder;
+        ConcurrentDictionary<string, LearningTransportQueueLengthTracker> trackers = new ConcurrentDictionary<string, LearningTransportQueueLengthTracker>(StringComparer.InvariantCultureIgnoreCase);
+
+        public Dictionary<EndpointInstanceId, int> SnapshotValues()
+        {
+            return (
+                from t in trackers.Values
```
Scooletz: Accessing `Values` or `Keys` in a `ConcurrentDictionary`, locks the whole dictionary. I'd suggest iterating over the whole dictionary `trackers` as enumerator is lock-free.

src/ServiceControl.LearningTransport/LearningTransportNativeQueueLengthProvider.cs
```diff
@@ -0,0 +1,88 @@
...
+            trackers.AddOrUpdate(queueName,
+                q =>
+                {
+                    var tracker = new LearningTransportQueueLengthTracker(rootFolder, q);
+                    tracker.Add(endpointInstanceId);
+                    return tracker;
+                },
+                (q, existing) =>
+                {
+                    existing.Add(endpointInstanceId);
```
Scooletz: Is there a possibility that another thread will obtain the tracker and will try to do something else with it at the same time? In other words, is the tracker itself thread-safe and should it be?

tmasternak:
Spike is over.

PR #121, opened 2018-05-03
Native Queue Length Infrastructure

mikeminutillo:
All the bits of https://github.com/Particular/ServiceControl.Monitoring/pull/118 that are not specific to SQL.

Introduces a Queue Length provider that can handle Queue Length related messages from clients and is responsible for updating the Queue Length Store. 

The Metric Report contains native queue lengths as recorded by the endpoint. The default Queue Length Provider simply writes those directly to the Queue Length Store.

The Endpoint Metadata Report contains a list of queues that the endpoint reads from. The default Queue Length Provider ignores these. Another provider might use these to determine which queues to monitor in a broker and how to map those back to the endpoints.

The Queue Length Store is broken down by endpoint and input queue. This allows us to handle competing consumers (where input queue and endpoint is the same) and sender-side distribution (where endpoint is the same but input queue differs). 

It is intended that each transport that is capable of centralized monitoring will provide its own custom SC transport assembly with a Queue Length Provider implementation in it.

All transports that prefer decentralized monitoring will work with the Metrics.ServiceControl plugin to send Metric Report data.


src/ServiceControl.Monitoring/QueueLength/DefaultQueueLengthProvider.cs
```diff
@@ -0,0 +1,40 @@
+﻿namespace ServiceControl.Monitoring.QueueLength
+{
+    using System.Threading.Tasks;
+    using Infrastructure;
+    using Messaging;
+    using NServiceBus.Metrics;
+
+    public class DefaultQueueLengthProvider : IProvideQueueLength
```
tmasternak: This class looks as if it was prepared for being a base class for any transport specific implementation. 
I think this might be a bit premature.

src/ServiceControl.Monitoring/Http/Diagrams/Aggregator.cs
```diff
@@ -22,27 +22,22 @@ internal static MonitoredValues AggregateTimings<T>(List<IntervalsStore<T>.Inter
...
         }
 
-        internal static MonitoredValues AggregateQueueLength<T>(List<IntervalsStore<T>.IntervalsBreakdown> intervals, HistoryPeriod period)
+        internal static MonitoredValues ToSumOfBreakdownAverages<T>(List<IntervalsStore<T>.IntervalsBreakdown> intervals, HistoryPeriod period)
         {
             Func<long, double> returnOneIfZero = x => x == 0 ? 1 : x;
 
             return new MonitoredValues
             {
-                // To calculate the average we count TotalMeasurements instead of the number of intervals.
```
tmasternak: @Scooletz this comment was removed as previously this piece of code was QL specific (method was called `AggreateQueueLength` instead of `ToSumOfBreakdownAverages`). I assumed that method rename will make it more clear what to expect.

tmasternak:
> NOTE: The When_querying_queue_length_data test is failing because the test endpoint doesn't handle the message it sends to itself. I'm not entire sure how this test ever worked.

@mikeminutillo fixed. I'm not sure how I missed this. Either way it's fixed now.

src/ServiceControl.Monitoring/EndpointFactory.cs
```diff
@@ -76,12 +82,32 @@ public static void MakeMetricsReceiver(EndpointConfiguration config, Settings se
...
 
-        static IContainer CreateContainer(Settings settings)
+        static Func<QueueLengthStore, IProvideQueueLength> QueueLengthProviderBuilder(string explicitConnectionStringValue, Type transportType)
+        {
+            return qls =>
+            {
+                var connectionString = explicitConnectionStringValue ?? ConfigurationManager.ConnectionStrings["NServiceBus/Transport"]?.ConnectionString;
+
+                var queueLengthProviderType = transportType.Assembly.GetTypes()
+                    .SingleOrDefault(p => typeof(IProvideQueueLength).IsAssignableFrom(p));
```
Scooletz: Just a thought: do want to fail on multiple implementations or should we pick up one and just log it? With the current approach, `SingleOrDefault` will throw on multiple ones.

src/ServiceControl.Monitoring/Http/Diagrams/Aggregator.cs
Scooletz: I retract my comment about removing the comment. Now, after looking at the implementation and the name of the method, it's a well defined function (a sum of averages) which is the last but one line in the LINQ statement

src/ServiceControl.Monitoring/QueueLength/DefaultQueueLengthProvider.cs
Scooletz: `virtual` + `protected` without inheritors? I agree with @tmasternak

tmasternak:
> One thing that worries me a little bit is the combined number of lines that we spent on aggregating values and presenting them via GET in Nancy module. On the other hand, this is a full-blown in-memory walking window aggregator. I can only imagine the LINQ statement describing it for EF.

@Scooletz I think this is a valid point. We are reaching the point at which I personally have trouble wrapping my head around the logic there. 

Nice proposal made by @mikeminutillo was to follow `composition-gateway` type of design. In such approach we would push parts of the module to each metric-type slice of monitoring instance. Queue length feature would attach it's aggregations to the response, timings feature its aggregations etc. The Nancy module would only setup response and iterate over such "attachers". That being said this deserve separate PR IMHO.

src/ServiceControl.Monitoring/EndpointFactory.cs
tmasternak: I can't think of scenario in which transport would have more than one implementation. I would go with fail fast approach and keep `SingelOrDefault`.

Scooletz:
@tmasternak @mikeminutillo I'm not authorize to merge it 😕

src/ServiceControl.Monitoring/QueueLength/DefaultQueueLengthProvider.cs
mikeminutillo: Fair enough. It was just transferred from the other PR. It _is_ intended to be a base class for any transport specific implementation. When I wrote it, it was an abstract base class and I created an MSMQ specific implementation over the top. Then I realized that the MSMQ implementation would have no code in it and that we don't need an MSMQ specific version.

mikeminutillo:
I also, am not authorized to merge.

PR #123, opened 2018-05-09
Legacy queue length logger fix

tmasternak:
## Overview
This PR fixes warn logging for the legacy queue length reports. It's supposed to log a warn once per endpoint `instanceId`. The implementation was based however on a dictionary that had per handler lifetime scope resulting in no de-duplication being done.

The fix proposed is to have a singleton of dedicated type that holds the reported legacy endpoints.

### DIFF
--- a/src/ServiceControl.Monitoring/QueueLength/LegacyQueueLengthReportHandler.cs
+++ b/src/ServiceControl.Monitoring/QueueLength/LegacyQueueLengthReportHandler.cs
@@ -2,6 +2,7 @@
 {
     using System;
     using System.Collections.Concurrent;
+    using System.Threading;
     using System.Threading.Tasks;
     using Infrastructure;
     using NServiceBus;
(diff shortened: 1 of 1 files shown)


mikeminutillo:
I wonder if we should just limit it to once per hour/day/something so that if the situation doesn't get resolved the warning keeps coming up.

We could also use it to attach a boolean flag to the viewmodel for each endpoint / instance so that Pulse can display a warning. It wouldn't take much additional work.

src/ServiceControl.Monitoring/QueueLength/LegacyQueueLengthReportHandler.cs
```diff
@@ -9,20 +10,47 @@
...
+            ConcurrentDictionary<string, string> regsteredInstances = new ConcurrentDictionary<string, string>();
+            DateTime lastCleanTime = DateTime.UtcNow;
+
+            public bool TryAdd(string id)
+            {
+                var now = DateTime.UtcNow;
+
+                if (lastCleanTime.Add(cleanInterval) < now)
+                {
+                    lastCleanTime = now;
```
Scooletz: I'm sure that because of the operations on the `registeredInstances` this won't happen, but still, I want to notice that a field is written without any lock. This means that depending on the impl of copying this struct, the assignment might not be atomic and one could see a partially written value.

src/ServiceControl.Monitoring/QueueLength/LegacyQueueLengthReportHandler.cs
```diff
@@ -9,20 +10,47 @@
...
             }
 
             return TaskEx.Completed;
         }
 
+        public class LegacyQueueLengthEndpoints
+        {
+            static TimeSpan cleanInterval = TimeSpan.FromHours(1);
+
+            ConcurrentDictionary<string, string> regsteredInstances = new ConcurrentDictionary<string, string>();
```
Scooletz: `regstered` should be probably `registered`

src/ServiceControl.Monitoring/QueueLength/LegacyQueueLengthReportHandler.cs
tmasternak: I've switched to `long` as the field type and added `Interlocked` for setting the value.

src/ServiceControl.Monitoring/QueueLength/LegacyQueueLengthReportHandler.cs
```diff
@@ -9,20 +11,47 @@
...
+            static long cleanIntervalTicks = TimeSpan.FromHours(1).Ticks;
+
+            ConcurrentDictionary<string, string> registeredInstances = new ConcurrentDictionary<string, string>();
+            long lastCleanTicks = DateTime.UtcNow.Ticks;
+
+            public bool TryAdd(string id)
+            {
+                var nowTicks = DateTime.UtcNow.Ticks;
+
+                if (Volatile.Read(ref lastCleanTicks) + cleanIntervalTicks < nowTicks)
```
Scooletz: What a beautiful usage of `Volatile.Read` ! 💯

PR #127, opened 2018-05-10
Custom Azure Storage Queues transport

mikeminutillo:
This is a necessary step before the native queue length provider can be added


Scooletz:
Could it be squashed and rebased? Because of the update from develop it shows RabbitMQ in the set of changes @tmasternak

tmasternak:
This has been done in #131

PR #133, opened 2018-05-30
[WIP] RabbitMQ native queue length

tmasternak:


bording:
> Am I handling connection recoverability in a correct way i.e. what happens if the connections goes down?

With automatic recovery enabled, it will reconnect if it loses the connection to the broker. However, if you try to use the channel and connection while in the reconnecting state, they'll throw exceptions.
 I'd recommend using the `ConnectionShutdown` and `RecoverySucceeded` Connection events to detect these states and stop/start your loop.

`ConnectionShutdown` has `ShutdownEventArgs` which can be used to determine why the connection is closing. You can see an example of that in the RabbitMQ transport message pump.

> I did not verify what happens with DefineQueuePasive if the queue for some reason does not exists. Do you know if the channel still usable or should I recreate it?

If the queue does not exist, you get an channel exception and the channel is closed. You'll have to catch an exception and create a new channel.


BTW, there is actually a `MessageCount` API on the channel that you could use instead of `QueueDeclarePassive`. At the protocol level doing the same thing as `QueueDeclarePassive`, but so it's basically a convenience method.

src/ServiceControl.Transports.RabbitMQ/QueueLengthProvider.cs
```diff
@@ -0,0 +1,142 @@
...
+        {
+            //RabbitMQ does not support endpoint level queue length reports
+        }
+
+        public Task Start()
+        {
+            var factory = new ConnectionFactory
+            {
+                Uri = connectionString,
+                AutomaticRecoveryEnabled = true,
```
bording: `AutomaticRecoveryEnabled` is true by default for the current client, so you don't need to specify it.

src/ServiceControl.Transports.RabbitMQ/QueueLengthProvider.cs
```diff
@@ -0,0 +1,142 @@
...
+                TopologyRecoveryEnabled = true
```
bording: This one is true by default as well, but I'd recommend setting it to false instead. You don't actually have anything here to recover, so there's no reason to have it enabled.

src/ServiceControl.Transports.RabbitMQ/QueueLengthProvider.cs
```diff
@@ -0,0 +1,142 @@
...
+                            DateTicks = nowTicks,
+                            Value = sizes.TryGetValue(endpointQueuePair.Value, out var size) ? size : 0
+                        }
+                    },
+                    endpointQueuePair.Key);
+        }
+
+        void FetchQueueLengths(IModel model)
+        {
+            //TODO: check what happens when non existing queue is querried
```
bording: As mentioned in comment, it throws an exception and the channel is closed.

src/ServiceControl.Transports.RabbitMQ/QueueLengthProvider.cs
```diff
@@ -0,0 +1,142 @@
...
+            foreach (var endpointQueuePair in endpointQueues)
+            {
+                try
+                {
+                    var declareOK = model.QueueDeclarePassive(endpointQueuePair.Value);
```
bording: As mentioned already, you can use `model.MessageCount` instead.

src/ServiceControl.Transports.RabbitMQ/QueueLengthProvider.cs
```diff
@@ -0,0 +1,142 @@
...
+        public void Process(EndpointInstanceId endpointInstanceId, TaggedLongValueOccurrence metricsReport)
+        {
+            //RabbitMQ does not support endpoint level queue length reports
+        }
+
+        public Task Start()
+        {
+            var factory = new ConnectionFactory
+            {
+                Uri = connectionString,
```
bording: What sort of connection string are you expecting here? The format described in https://docs.particular.net/transports/rabbitmq/connection-settings is something we created, and is not what the client is expecting.

The `Uri` should be something like `amqp://rabbit.mshome.net:5672`. 

For consistency, it might make sense to bring in the code from the transport to support the same type of connection string.

tmasternak:
@bording thanks for the input!

I've added the connection string parsing logic (taken from the transport) and moved the connection, model failure handling to a separate class `QueryExecutor`. I did not go with the callback - checking the status flag looks good enough to me. Thoughts?

src/ServiceControl.Transports.RabbitMQ/QueueLengthProvider.cs
```diff
@@ -0,0 +1,177 @@
...
+            var nowTicks = DateTime.UtcNow.Ticks;
+
+            foreach (var endpointQueuePair in endpointQueues)
+                store.Store(
+                    new[]
+                    {
+                        new RawMessage.Entry
+                        {
+                            DateTicks = nowTicks,
+                            Value = sizes.TryGetValue(endpointQueuePair.Value, out var size) ? size : 0
```
mikeminutillo: Is it correct to report 0 where we have no measurement?

src/ServiceControl.Transports.RabbitMQ/QueueLengthProvider.cs
```diff
@@ -0,0 +1,177 @@
...
+
+                    action(model);
+                }
+                catch (Exception e)
+                {
+                    Logger.Warn("Error querying queue length.", e);
+                }
+            }
+
+            TimeSpan ReconnectionDelay = TimeSpan.FromSeconds(5);
```
mikeminutillo: This is _forever_. What if the reconnection happens just after we checked the flag?

tmasternak: This is based on the fact that 5 sec is the default [timeout](https://www.rabbitmq.com/dotnet-api-guide.html#connection-recovery) for the built-in rabbitmq client library recoverability.

src/ServiceControl.Transports.RabbitMQ/ConnectionConfiguration.cs
```diff
@@ -0,0 +1,194 @@
...
+                {
+                    invalidOptionsMessage.AppendLine($"'{parts[1]}' is not a valid Int32 value for the port in the 'host' connection string option.");
+                }
+            }
+            else
+            {
+                invalidOptionsMessage.AppendLine("Invalid connection string. 'host' value must be supplied. e.g: \"host=myServer\"");
+            }
+
+            var nsbVersion = FileVersionInfo.GetVersionInfo(typeof(Endpoint).Assembly.Location);
```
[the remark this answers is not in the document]
tmasternak: Transport specific packages already have (or are predicted to have) transport reference and as a result core reference so I think this is unavoidable.

src/ServiceControl.Transports.RabbitMQ/QueueLengthProvider.cs
tmasternak: Actually the default used by RabbitMQ client is 10 sec. See @bording  [comment](https://github.com/Particular/ServiceControl.Monitoring/pull/133#discussion_r193945454)

src/ServiceControl.Transports.RabbitMQ/QueueLengthProvider.cs
tmasternak: This is a good point.

tmasternak:
@bording can you have a look?

src/ServiceControl.Transports.RabbitMQ/QueueLengthProvider.cs
bording: 10 seconds is the default that the transport uses when creating a `ConnectionFactory`, but the client's default value if the `NetworkRecoveryInterval` property isn't modified is 5 seconds.

PR #134, opened 2018-06-06
Native Queue Length - Learning Transport

mikeminutillo:


src/ServiceControl.LearningTransport/QueueLengthProvider.cs
```diff
@@ -0,0 +1,121 @@
...
+                }
+            });
+
+            return Task.FromResult(0);
+        }
+
+        void UpdateStore(ILookup<string, long?> queueLengths)
+        {
+            var now = DateTime.UtcNow.Ticks;
+            foreach (var instance in endpointsHash.Keys)
```
Scooletz: Accessing keys, locks the dictionary. Please use enumerator over the dictionary and get instance from `kvp.Key`

Scooletz:
If I'm the one to review it, it looks good to me 👍

tmasternak:
@mikeminutillo in order to release it we still need to tweak the SCMU right?

mikeminutillo:
Discussed with @tmasternak over slack. We probably don't want to release support for the learning transport in SCMU until ServiceControl-prime can do that as well.

mikeminutillo:
With a smoke test in place, this is ready to merge

PR #137, opened 2018-06-08
SQS Queue Length Provider

Scooletz:
This PR provides SQS Queue Length Provider


Scooletz:
@danielmarbach If you could be so kind and review this PR of mine, I'd be more than happy 😉

src/ServiceControl.Transports.AmazonSQS/QueueLengthProvider.cs
```diff
@@ -0,0 +1,132 @@
...
+                        {
+                            await FetchQueueSizes(cache, client).ConfigureAwait(false);
+
+                            UpdateQueueLengthStore();
+
+                            await Task.Delay(QueryDelayInterval);
+                        }
+                        catch (Exception e)
+                        {
+                            Logger.Error("Error querying sql queue sizes.", e);
```
danielmarbach: `Error querying sql queue sizes.` should be `SQS` and probably give a hint about what went wrong?

src/ServiceControl.Transports.AmazonSQS/QueueLengthProvider.cs
```diff
@@ -0,0 +1,132 @@
...
+
+                    while (!stop.Token.IsCancellationRequested)
+                    {
+                        try
+                        {
+                            await FetchQueueSizes(cache, client).ConfigureAwait(false);
+
+                            UpdateQueueLengthStore();
+
+                            await Task.Delay(QueryDelayInterval);
```
danielmarbach: `ConfigureAwait(false)`

src/ServiceControl.Transports.AmazonSQS/QueueLengthProvider.cs
```diff
@@ -0,0 +1,132 @@
...
+                    new[]{ new RawMessage.Entry
+                    {
+                        DateTicks = nowTicks,
+                        Value = sizes.TryGetValue(tableNamePair.Value, out var size) ? size : 0
+                    }},
+                    tableNamePair.Key);
+            }
+        }
+
+        Task FetchQueueSizes(QueueUrlCache cache, IAmazonSQS client) => Task.WhenAll(sizes.Select(kvp => FetchLength(kvp.Key, client, cache)));
```
danielmarbach: I would favor a list of tasks (the list could be owned by the loop) that get cleared and a foreach loop here to avoid allocating too much but maybe not worth optimizing?

src/ServiceControl.Transports.AmazonSQS/QueueLengthProvider.cs
```diff
@@ -0,0 +1,132 @@
...
+
+        async Task FetchLength(string queue, IAmazonSQS client, QueueUrlCache cache)
+        {
+            try
+            {
+                var attReq = new GetQueueAttributesRequest
```
danielmarbach: Would it make sense to get the queue url and create the `GetQueueAttributesRequest` request (yes it can be reused) when we see it in the `Process` method or should the process be as shortlived as possible? If that is the case maybe we can tweak it a bit so that we don't need to recreate the `GetQueueAttributesRequest` over and over again given the queues should be fairly stable over time? So maybe instead of caching the URL we cache the whole request object?

src/ServiceControl.Transports.AmazonSQS/QueueLengthProvider.cs
```diff
@@ -0,0 +1,132 @@
...
+                {
+                    QueueUrl = await cache.GetQueueUrl(queue).ConfigureAwait(false)
+                };
+                attReq.AttributeNames.Add("ApproximateNumberOfMessages");
+                var response = await client.GetQueueAttributesAsync(attReq).ConfigureAwait(false);
+                sizes[queue] = response.ApproximateNumberOfMessages;
```
danielmarbach: What are the caveats here of using the indexer vs the AddOrUpdate methods?

Scooletz: Indexer uses a blunt fast way of adding or updating by 

```c#
TryAddInternal(key, value, true, true, out _);
```

`AddOrUpdate` uses merge function (and a factory function), a bit heavier.

src/ServiceControl.Transports.AmazonSQS/QueueLengthProvider.cs
Scooletz: Will do

Scooletz:
@danielmarbach amended.

src/ServiceControl.Transports.AmazonSQS/QueueAttributesRequestCache.cs
```diff
@@ -0,0 +1,43 @@
...
+
+    class QueueAttributesRequestCache
+    {
+        public QueueAttributesRequestCache(IAmazonSQS sqsClient)
+        {
+            cache = new ConcurrentDictionary<string, GetQueueAttributesRequest>();
+            this.sqsClient = sqsClient;
+        }
+
+        public async Task<GetQueueAttributesRequest> GetQueueAttributesRequest(string queueName)
```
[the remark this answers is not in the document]
Scooletz: Yes it does.

Scooletz:
@danielmarbach `ConfigureAwait(false)` ? Done

PR #147, opened 2018-10-02
ASBS transport support

mauroservienti:
## Overview
This PR introduces ASBS as a supported transport. It also adds support for 2 topologies in RabbitMQ.

Changes needed in ServiceControl installer are handled in https://github.com/Particular/ServiceControl/pull/1443.

Open issues/questions:
- ~all in Monitoring is .NET 4.5.2 however ASBS is only .NET Standard 2.0, meaning the minimum full fx version is 4.6.1. Is this a problem? Shall we upgrade everything? SC is already at 4.6.1~ This will be handled in a separate PR.
- ~there is no concept of `NamespaceManager` and Queue Length in the new MS SDK, who should I ping to understand how to build the native queue length provider?~ We are good with the current approach.
- ~Is it OK to have 2 transports (the ASB Legacy ones) in the same package? I bundled the 2 legacy topology in the same AzureServiceBus package~ Yes this is fine. We now have 2 transports in one assembly


mauroservienti:
Raised https://github.com/Particular/ServiceControl/pull/1443 to start tracking changes that need to be applied in SC to release thew new monitoring stuff

SeanFeldman:
> there is no concept of NamespaceManager and Queue Length in the new MS SDK

It's now called `ManagementClient`.

mauroservienti:
Thanks @SeanFeldman

src/ServiceControl.Transports.AzureServiceBusStandard/QueueLengthProvider.cs
```diff
@@ -0,0 +1,114 @@
...
+            {
+                var queue = queueData[mapping.Value].FirstOrDefault();
+                if (queue != null)
+                {
+                    var entries = new[]
+                    {
+                        new RawMessage.Entry
+                        {
+                            DateTicks = timestamp,
+                            Value = (await managementClient.GetQueueRuntimeInfoAsync( queue.Path )).MessageCountDetails.ActiveMessageCount
```
tmasternak: We might want to consider chunking the queueData. In would enable us to make sure we don't make more than x requests per delay-interval. E.g. we ask at most about 20 queues each interval. I would consider that in the context of the amount of requests that are issued to the broker. @SeanFeldman thoughts?

src/ServiceControl.Transports.AzureServiceBusStandard/QueueLengthProvider.cs
```diff
@@ -0,0 +1,114 @@
...
+                else
+                {
+                    Logger.DebugFormat("Endpoint {0} ({1}): no queue length data found for queue {2}", mapping.Key.EndpointName, mapping.Key.InstanceName ?? mapping.Key.InstanceId, mapping.Value);
+                }
+            }
+        }
+
+        public Task Stop()
+        {
+            stop.Cancel();
```
SeanFeldman: `ManagementClient` needs to be closed as well.

src/ServiceControl.Transports.AzureServiceBusStandard/QueueLengthProvider.cs
SeanFeldman: It depends on the number of queues that have to be queried. But keep in mind that this is a different API, that has nothing to do with message sending/receiving, and AFAIK is not a subject to throttling.

Generally, the messaging team doesn't recommend this API for usage such as this one, but alternative is to go through AAD, which would not work with connection string.s

src/ServiceControl.Transports.AzureServiceBusStandard/QueueLengthProvider.cs
```diff
@@ -0,0 +1,114 @@
...
+
+                        var queues = await managementClient.GetQueuesAsync().ConfigureAwait(false);
+                        var lookup = queues.ToLookup(x => x.Path, StringComparer.InvariantCultureIgnoreCase);
+
+                        Logger.DebugFormat("Retrieved details of {0} queues", lookup.Count);
+
+                        await UpdateQueueLengthStore(lookup);
+
+                        Logger.Debug("Waiting for next interval");
+                        await Task.Delay(QueryDelayInterval).ConfigureAwait(false);
```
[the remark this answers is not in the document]
mauroservienti: done

src/ServiceControl.Transports.AzureServiceBusStandard/QueueLengthProvider.cs
mauroservienti: done

mauroservienti:
Added smoke tests in https://github.com/Particular/ServiceControl.Monitoring/pull/147/commits/97a73c781beb1abc3c815e76637aa3cf1e681404 by copying the ones from ASB, can someone please have a look at the default server implementation?

src/ServiceControl.Transports.AzureServiceBusStandard/QueueLengthProvider.cs
tmasternak: In that case, I say looks-good-to-me. Thx @SeanFeldman !

mauroservienti:
@tmasternak I applied the changes you requested. Can you do another review?

mauroservienti:
rebased on `develop` and changed the base branch for the PR to `develop`

src/ServiceControl.Monitoring.SmokeTests.ASB.EndpointOrientedTopology/app.config
```diff
@@ -7,9 +7,43 @@
...
+        <assemblyIdentity name="Microsoft.IdentityModel.Clients.ActiveDirectory" publicKeyToken="31bf3856ad364e35" culture="neutral" />
+        <bindingRedirect oldVersion="0.0.0.0-3.19.8.16603" newVersion="3.19.8.16603" />
+      </dependentAssembly>
+      <dependentAssembly>
+        <assemblyIdentity name="Microsoft.IdentityModel.Clients.ActiveDirectory.Platform" publicKeyToken="31bf3856ad364e35" culture="neutral" />
+        <bindingRedirect oldVersion="0.0.0.0-3.19.8.16603" newVersion="3.19.8.16603" />
       </dependentAssembly>
     </assemblyBinding>
   </runtime>
+  <system.serviceModel>
```
andreasohlund: Is all this stuff needed?

tmasternak:
@bording this is ready for the initial review.

src/Packaging.NServiceBus.AzureServiceBus/Packaging.NServiceBus.AzureServiceBus.csproj
```diff
@@ -39,22 +38,22 @@
...
+  </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\ServiceControl.Transports.AzureServiceBus\ServiceControl.Transports.AzureServiceBus.csproj">
-      <Project>{66ca123a-7e48-494f-a3d3-ded5dd664621}</Project>
+      <Project>{a5a18ce5-2065-4e9a-a8a3-e55cabb50407}</Project>
       <Name>ServiceControl.Transports.AzureServiceBus</Name>
     </ProjectReference>
   </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
-  <Target Name="AfterBuild">
```
bording: Why was this removed, but added in the new Packaging.NServiceBus.AzureServiceBusStandard.csproj file?

src/ServiceControl.Transports.AzureServiceBusStandard/QueueLengthProvider.cs
```diff
@@ -0,0 +1,114 @@
+namespace ServiceControl.Transports.AzureServiceBus
```
bording: Namespace doesn't look correct

src/ServiceControl.Transports.AzureServiceBus/ServiceControl.Transports.AzureServiceBus.csproj
```diff
@@ -5,6 +5,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="NServiceBus" Version="7.1.4" />
```
bording: Why was this added?

src/ServiceControl.Transports.AzureServiceBusStandard/ServiceControl.Transports.AzureServiceBusStandard.csproj
```diff
@@ -0,0 +1,15 @@
...
+
+  <PropertyGroup>
+    <TargetFramework>net461</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="NServiceBus.Transport.AzureServiceBus" Version="1.0.0-alpha0164" />
+  </ItemGroup>
+
+  <ItemGroup>
```
bording: Nitpick, but I prefer to see project references first, then package references.

src/ServiceControl.Transports.RabbitMQ/ConventialTopologyRabbitMQTransport.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace ServiceControl.Transports.RabbitMQ
+{
+    using NServiceBus;
+    using NServiceBus.Settings;
+    using NServiceBus.Transport;
+
+    public class ConventialTopologyRabbitMQTransport : RabbitMQTransport
```
bording: Should be `ConventionalRoutingTopologyRabbitMQTransport`

src/ServiceControl.Transports.RabbitMQ/DirectTopologyRabbitMQTransport.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace ServiceControl.Transports.RabbitMQ
+{
+    using NServiceBus;
+    using NServiceBus.Settings;
+    using NServiceBus.Transport;
+
+    public class DirectTopologyRabbitMQTransport : RabbitMQTransport
```
bording: Should be `DirectRoutingTopologyRabbitMQTransport`

bording:
There's also a lot of inconsistency in how we're distingushing between the the two transports from a naming perspective.

The CustomTransportConfiguration projects use AzureServiceBus / AzureServiceBusStandard, but the SmokeTests are ASB / ASBS.

There's also Transports\AzureServiceBus and Transports\NetStandardAzureServiceBus

I think we should be standardizing on how we refer to these. From an end-user UI standpoint we're going with "AzureServiceBus Legacy" and "AzureServiceBus", right? I think we should align all the naming in that way then.

tmasternak:
@bording Regarding naming conventions 

> The CustomTransportConfiguration projects use AzureServiceBus / AzureServiceBusStandard, but the SmokeTests are ASB / ASBS.

I propose to tackle it in #151 and call it out of scope for this one.

> There's also Transports\AzureServiceBus and Transports\NetStandardAzureServiceBus

This is caused by logic in the SCMU that deals with package un-zipping. The `zipFolderNameToExtract` argument in [FileUtils](https://github.com/Particular/ServiceControl/blob/develop/src/ServiceControlInstaller.Engine/FileSystem/FileUtils.cs#L89) is the zip subfolder name e.g. `NetStandardAzureServiceBus`. The un-zipping logic does matching with `string.StartsWith` which causes the logic to blow when one folder is a prefix of the other. I'm not sure why we have such logic in place but this is the reason for such naming. AFAIK ServiceControl follows this naming as well.

> I think we should be standardizing on how we refer to these. From an end-user UI standpoint we're going with "AzureServiceBus Legacy" and "AzureServiceBus", right? I think we should align all the naming in that way then.

I'm not sure we made a decision yet /cc: @andreasohlund. If not I propose to continue with current naming and make the change in this PR or another one when we agree on the names.

andreasohlund:
>I'm not sure we made a decision yet /cc: @andreasohlund. If not I propose to continue with current naming and make the change in this PR or another one when we agree on the names.

Lets agree on the names and then make the needed adjustments?

I propose the following display names:

* AzureServiceBus Legacy - Endpoint oriented topology
* AzureServiceBus Legacy - Forwarding topology
* AzureServiceBus

src/ServiceControl.Transports.AzureServiceBusStandard/ServiceControl.Transports.AzureServiceBusStandard.csproj
tmasternak: done

tmasternak:
@bording regarding `<AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>`. First of all thanks for pointing that out :).

Secondly, it almost works out of the box ;). There were two things needed:
 * I had to add full target definition as in this [suggestion](https://github.com/dotnet/sdk/issues/1351#issuecomment-309634243). Looks that this is a problem for the old csproj file format only,
 * For two projects I had to add explicit `PackageReference` to `Newtonsoft.Json` otherwise the redirects were not generated for those assemblies. Once more looks like this should go away when we migrate to new csproj format.

bording:
> This is caused by logic in the SCMU that deals with package un-zipping. The zipFolderNameToExtract argument in FileUtils is the zip subfolder name e.g. NetStandardAzureServiceBus. The un-zipping logic does matching with string.StartsWith which causes the logic to blow when one folder is a prefix of the other. I'm not sure why we have such logic in place but this is the reason for such naming. AFAIK ServiceControl follows this naming as well.

@tmasternak In that case, perhaps we could go with `LegacyAzureServiceBus` and `AzureServiceBus`?

andreasohlund:
The idea is to agree on the naming here and then change SC accordingly do lets go with your proposed corrections?

tmasternak:
@bording  after syncing with @andreasohlund  we propose:
> Azure Service Bus Legacy - Endpoint-oriented topology
> Azure Service Bus Legacy - Forwarding topology
> Azure Service Bus

bording:
> Azure Service Bus Legacy - Endpoint-oriented topology
> Azure Service Bus Legacy - Forwarding topology
> Azure Service Bus

I like those, but I also saw a suggestion from @SeanFeldman that's worth considering:

> Azure Service Bus - Endpoint-oriented topology (Legacy)
> Azure Service Bus - Forwarding topology (Legacy)
> Azure Service Bus

It's really just about what emphasis we want on the "Legacy" aspect of the name.

Also, since I assume we can control the ordering of the entries in the settings dropdown, we should probably make sure the "Azure Service Bus" entry comes before the legacy entries.

src/ServiceControl.Transports.AzureServiceBusStandard/ServiceControl.Transports.AzureServiceBusStandard.csproj
tmasternak: done :)

tmasternak:
One note about the naming. In the current names `AzureServcieBus` means `Azur Service Bus - Forwarding topology (Legacy)` this might be a bit confusing for someone using PowerShell as `Azure Service Bus` will mean netstandard version. I propose to discuss that on the sync meeting.

tmasternak:
This is possible options for halding legacy transprot selection in SCMU
image

tmasternak:
After a short discussion with @danielmarbach he proposed to have two dropdowns for transports and legacy transports.

andreasohlund:
I like the idea of making it harder for  users to find the legacy ones so
:+1:

tmasternak:
@bording I've aligned the names of smoke test projects. Can you have a look?

Regarding other name aligning I propose to:
 * not go with the "Legacy" in the transport name as we removed that an introduced `IsLegacy` flat in SC transport definition. See: https://github.com/Particular/ServiceControl/pull/1469/files#diff-87d7b568504bc8a8c4d1afb94aad05a1R23
 * not to change the transport folder name, currently, we can have the same name as the name of the project itself and I would call it out of scope for this issue to fix the installer logic.

bording:
I've pushed up a few changes/fixes for things in the new commits since the last time I looked at the PR.

Side note, the Packaging.NServiceBus.Learning package/folder should probably be renamed to be "LearningTransport" to be consistent with the other projects.

So, at this point the other projects are at least consistently named within this solution, but we do have a consistency problem across other repos now. For example, we've gone with "MSMQ" here, but in other places we have projects that go with "Msmq" instead.

Of course, that sort of thing is out of scope for this PR, but we should note it somewhere for the future.

bording:
> not go with the "Legacy" in the transport name as we removed that an introduced IsLegacy flat in SC transport definition. See: https://github.com/Particular/ServiceControl/pull/1469/files#diff-87d7b568504bc8a8c4d1afb94aad05a1R23

I'm still not sure I'm happy with the current naming in this PR. "AzureServiceBusStandard" doesn't feel right to me given that we are actively trying to remove the "ASBS" naming everywhere.

That also goes for the "NetStandardAzureServiceBus" name for the packaging as well.

tmasternak:
> Side note, the Packaging.NServiceBus.Learning package/folder should probably be renamed to be "LearningTransport" to be consistent with the other projects. Also, we're still calling it "Learning" in the resulting zip file.

I've fixed those.

> So, at this point the other projects are at least consistently named within this solution, but we do have a consistency problem across other repos now. For example, we've gone with "MSMQ" here, but in other places we have projects that go with "Msmq" instead.

I agree it's out of scope.


> I'm still not sure I'm happy with the current naming in this PR. "AzureServiceBusStandard" doesn't feel right to me given that we are actively trying to remove the "ASBS" naming everywhere.

I think the problem is that we can have only one transport package named AzureServiceBus. So we can either say with the current namings or we could rename the old one to something like AzureServiceBusLegacy. Personally, I don't see much value in renaming so propose to keep to current names.

> That also goes for the "NetStandardAzureServiceBus" name for the packaging as well.

As previously stated this is caused by installer quirk and I think it's out of the scope of this PR.

SeanFeldman:
Well, if there was not enough confusion already, this https://github.com/Particular/ServiceControl.Monitoring/pull/147#issuecomment-433135838 has added a bit more 🙂Sorry.

I've scheduled a call for Friday 7AM MST that will include TF and @Particular/dotnetcore-launch-squad members to review and discuss what needs to happen. Apology for delaying your process, though it should help to get this work back on track in a better shape. (🤞)

andreasohlund:
With @tmasternak on vacation we need either @Scooletz or @mikeminutillo to merge this, sorry about this we should have planned better :(

PR #148, opened 2018-10-03
Query delay as first step in sql ql reporter

tmasternak:


Scooletz:
@tmasternak Could you add some description to this change? 😋

PR #160, opened 2018-11-16
Make sure netstandard facades and assembly redirects is shipped

andreasohlund:
To make sure we get facade assemblies and assembly redirects properly generated for ASBS to run on net461

Closes https://github.com/Particular/ServiceControl.Monitoring/issues/161


andreasohlund:
@tmasternak talked to @bording and we think this is the safest move to make sure we have the redirects in the config file and also the facade assemblies present.

Thoughts? (still need to make sure we pick all this up when packaging)

andreasohlund:
@bording like we talked about yesterday. Please take a look?

bording:
@andreasohlund I pushed up a small tweak because you don't actually need the empty App.config file. I assume the App-Dev.config file is being kept to you could just temporarily rename it if you need it?

Overall, this seems to get the zip file generally looking correct. Am I correct in thinking that the transport folder gets copied on top of the main folder? That seems like how it must work.


There does still seem to be some general weirdness around what is getting included in the transport folders, because the the bin content of the individual packaging projects don't match the content in the custom transport projects in ways I don't really understand.

src/ServiceControl.Monitoring/ServiceControl.Monitoring.csproj
```diff
@@ -214,6 +214,12 @@
   <ItemGroup>
     <Analyzer Include="..\packages\NServiceBus.7.1.4\analyzers\dotnet\cs\NServiceBus.Core.Analyzer.dll" />
   </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\DummyNetStandard20Ref\DummyNetStandard20Ref.csproj">
+      <Project>{6b406087-d062-4619-a56d-538b242f06ac}</Project>
+      <Name>DummyNetStandard20Ref</Name>
```
SeanFeldman: Does this mean `DummyNetStandard20Ref.dll` will show up among the files?
In case the answer is "yes", I'd only suggest to replace "dummy" with something more meaningful/descriptive. In case the answer is "no", call me dummy and ignore the comment 🙂

tmasternak:
> There does still seem to be some general weirdness around what is getting included in the transport folders, because the the bin content of the individual packaging projects don't match the content in the custom transport projects in ways I don't really understand.

@bording I'm pretty sure there is some skip list for transport specific packages that is why the folder for transport and genral set of assemblies set differ.

bording:
> @bording I'm pretty sure there is some skip list for transport specific packages that is why the folder for transport and genral set of assemblies set differ.

I saw that, but even when I disabled it, it still doesn't account for the differences I'm seeing.

PR #163, opened 2018-11-19
Bump patch version of ASB and ASBS

danielmarbach:
I decided not to bump to ASB 9.x but just to 8.0.6 (latest patch version) because I figured for some reason we are still using 8.x which I don't know TBH

### DIFF
--- a/src/ServiceControl.Monitoring.SmokeTests.LegacyAzureServiceBus/packages.config
+++ b/src/ServiceControl.Monitoring.SmokeTests.LegacyAzureServiceBus/packages.config
@@ -5,7 +5,7 @@
   <package id="Newtonsoft.Json" version="11.0.2" targetFramework="net46" />
   <package id="NServiceBus" version="7.1.4" targetFramework="net46" />
   <package id="NServiceBus.AcceptanceTesting" version="7.1.4" targetFramework="net46" />
-  <package id="NServiceBus.Azure.Transports.WindowsAzureServiceBus" version="8.0.6" targetFramework="net461" />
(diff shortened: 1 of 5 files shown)


src/ServiceControl.Monitoring.SmokeTests.LegacyAzureServiceBus/packages.config
```diff
@@ -5,7 +5,7 @@
   <package id="Newtonsoft.Json" version="11.0.2" targetFramework="net46" />
   <package id="NServiceBus" version="7.1.4" targetFramework="net46" />
   <package id="NServiceBus.AcceptanceTesting" version="7.1.4" targetFramework="net46" />
-  <package id="NServiceBus.Azure.Transports.WindowsAzureServiceBus" version="8.0.3" targetFramework="net46" />
+  <package id="NServiceBus.Azure.Transports.WindowsAzureServiceBus" version="8.0.6" targetFramework="net461" />
```
andreasohlund: Out of curiosity, what caused the bump from net46 to net461? Was that something we had changed in `NServiceBus.Azure.Transports.WindowsAzureServiceBus` ?

danielmarbach: That is a good question, don't know. I think it was the tooling of VS. Downgrading and see whether it works

andreasohlund: Sounds reasonable, the project is targeting 461 so I guess references you touch would get updated?

https://github.com/Particular/ServiceControl.Monitoring/blob/develop/src/ServiceControl.Transports.LegacyAzureServiceBus/ServiceControl.Transports.LegacyAzureServiceBus.csproj#L4

PR #172, opened 2018-12-18
Upgrade ASB and enable migration mode on endpoint oriented topology

danielmarbach:


src/ServiceControl.Monitoring.SmokeTests.LegacyAzureServiceBus/packages.config
```diff
@@ -5,11 +5,11 @@
...
   <package id="NServiceBus.AcceptanceTesting" version="7.1.4" targetFramework="net46" />
-  <package id="NServiceBus.Azure.Transports.WindowsAzureServiceBus" version="8.0.6" targetFramework="net46" />
+  <package id="NServiceBus.Azure.Transports.WindowsAzureServiceBus" version="9.1.0" targetFramework="net461" />
   <package id="NServiceBus.Metrics" version="3.0.0" targetFramework="net46" />
   <package id="NServiceBus.Metrics.ServiceControl" version="3.0.2" targetFramework="net46" />
   <package id="NServiceBus.Newtonsoft.Json" version="2.1.0" targetFramework="net46" />
   <package id="NUnit" version="3.10.1" targetFramework="net46" />
   <package id="System.IdentityModel.Tokens.Jwt" version="4.0.4.403061554" targetFramework="net46" />
-  <package id="WindowsAzure.ServiceBus" version="4.1.11" targetFramework="net46" />
+  <package id="WindowsAzure.ServiceBus" version="5.0.2" targetFramework="net461" />
```
SeanFeldman: Would this require a new major?
We're using a different .NET Framework here

danielmarbach: No we are not. The target FW is 4.6.1 already. It is nuget who upgrades that when we touch the file. Still good spot. We can leave it as is

SeanFeldman: I think ASB client will require net461, so if Monitoring is already on 4.6.1, shouldn't be an issue. Thanks.

PR #174, opened 2018-12-19
SQS Transport to accept EC2-Profile as  Credentials  Source

danielmarbach:
Allow SQS Transport customization to accept EC2 Profile ( Instance Role ) as a credential source.  Currently, the transport **requires** API Keys to be passed in connection string. Which is less secure and introduces ops complexity around key rotation.


danielmarbach:
See also https://github.com/Particular/ServiceControl/pull/1544

PR #190, opened 2019-02-15
Fix excessive queue length querying in ASQ

tmasternak:
### Description
Before this fix the `size` collection in the `NativeQueueLength` would grow one entry for each monitoring report received. This, in turn, resulted in excessive storage account usage due to the growing number of queue length queries.

### Fix
This PR introduces a dedicated class that stores all queue length related variables. This enables having only a single collection for queue length data which helps ensuring that the state of the `NativeQueueLength` does not grow indefinitely.

### DIFF
--- a/src/ServiceControl.Transports.AzureStorageQueues/QueueLengthProvider.cs
+++ b/src/ServiceControl.Transports.AzureStorageQueues/QueueLengthProvider.cs
@@ -16,9 +16,8 @@
 
     public class QueueLengthProvider : IProvideQueueLength
     {
-        ConcurrentDictionary<EndpointInputQueue, CloudQueue> queues = new ConcurrentDictionary<EndpointInputQueue, CloudQueue>();
-        ConcurrentDictionary<CloudQueue, int> sizes = new ConcurrentDictionary<CloudQueue, int>();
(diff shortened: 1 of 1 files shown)


src/ServiceControl.Transports.AzureStorageQueues/QueueLengthProvider.cs
```diff
@@ -38,19 +37,23 @@ public void Process(EndpointInstanceId endpointInstanceId, EndpointMetadataRepor
...
-            queues.AddOrUpdate(endpointInputQueue, _ => queue, (_, currentQueue) =>
+            var queueLength = new QueueLengthValue
             {
-                if (currentQueue.Name.Equals(queue.Name) == false)
+                QueueName = queueName,
+                Length = 0,
+                QueueReference = queueClient.GetQueueReference(queueName)
+            };
+
+            queueLengths.AddOrUpdate(endpointInputQueue, _ => queueLength, (_, currentLength) =>
```
SeanFeldman: I find this line confusing because of multiple usage of `_`. If it means "ignore" it's ok, but just confusing visually.

SeanFeldman:
Given I don't entirely know what Monitoring is supposed to do and what was it doing that cause the issue, could we chat first? Thanks.

PR #195, opened 2019-02-25
Remove API quota to allow serving large collections of message types

tmasternak:
## Overview
Some of our clients have a lot of message types flowing through single logic instances e.g. 600+. This PR removes quota at Nacy level to make sure that big responses are not truncated and can be properly served to the ServicePulse instances.


tmasternak:
@mikeminutillo good questions. This is one part of the solution, the other is changes to [SP](https://github.com/Particular/ServicePulse/pull/719). On the SP side, we do a browser side pagination. We testest that setup with 800 message types per endpoint and the performance is acceptable.

